PR #4, opened 2015-07-21
allow the bridge to find Python 2 packages from ROS 1

dirk-thomas:
Since all ROS 2 code uses Python 3 and the ROS 1 packages are only available in Python 2 this needs some weird way to include `catkin_pkg` and `rospkg`.

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -11,8 +11,47 @@ import rosidl_parser
 # ROS 1 imports
 import genmsg
 import genmsg.msg_loader
+
+# import catkin_pkg and rospkg which are required by rosmsg
+# and likely only available for Python 2
+try:
+    import catkin_pkg
+except ImportError:
+    from importlib.machinery import SourceFileLoader
+    import subprocess
+    for python_executable in ['python2', 'python2.7']:
+        try:
+            catkin_pkg_path = subprocess.check_output(
+                [python_executable, '-c', 'import catkin_pkg; print(catkin_pkg.__file__)'])
+        except subprocess.CalledProcessError:
+            continue
+        catkin_pkg_path = catkin_pkg_path.decode().strip()
+        if catkin_pkg_path.endswith('.pyc'):
+            catkin_pkg_path = catkin_pkg_path[:-1]
+        catkin_pkg = SourceFileLoader('catkin_pkg', catkin_pkg_path).load_module()
+    if not catkin_pkg:
+        raise
+
+try:
+    import rospkg
+except ImportError:
+    from importlib.machinery import SourceFileLoader
+    import subprocess
+    for python_executable in ['python2', 'python2.7']:
+        try:
+            rospkg_path = subprocess.check_output(
+                [python_executable, '-c', 'import rospkg; print(rospkg.__file__)'])
+        except subprocess.CalledProcessError:
+            continue
+        rospkg_path = rospkg_path.decode().strip()
+        if rospkg_path.endswith('.pyc'):
+            rospkg_path = rospkg_path[:-1]
+        rospkg = SourceFileLoader('rospkg', rospkg_path).load_module()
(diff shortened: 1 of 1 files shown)


ros1_bridge/__init__.py
```diff
@@ -11,8 +11,33 @@
...
 import genmsg.msg_loader
+
+# import catkin_pkg and rospkg which are required by rosmsg
+# and likely only available for Python 2
+try:
+    import catkin_pkg
+except ImportError:
+    from importlib.machinery import SourceFileLoader
+    from subprocess import check_output
+    catkin_pkg_path = check_output(['python2', '-c', 'import catkin_pkg; print(catkin_pkg.__file__)'])
```
wjwwood: Unfortunately, I think using just `python2` here is a weak assumption, on my machine:

```
% python2
zsh: command not found: python2
```

I do however have python2.7 and python which point to a python2.x interpreter. Maybe that's ok since this is just a recovery behavior, but I wouldn't assume this will work outside of Ubuntu like distros.

wjwwood:
+1, other than comment.

ros1_bridge/__init__.py
dirk-thomas: Following the Python recommendations that is the only thing I could imagine. `python` might point to either 2 or 3.

Are the ROS 1 `catkin_pkg` and `rospkg` packages available when you use Python 3 on OS X?

wjwwood: They would be if I installed them for Python3, but no the Python3 interpreter does not get the Python2 site-packages by default.

dirk-thomas:
I will update the patch to also try `python2.7` when `python2` fails to invoke.

dirk-thomas:
Updated.

@wjwwood Can you please try to build this package on OS X.

esteve:
+1

dirk-thomas:
I will merge this since it is necessary for building the bridge on linux anyway.

wjwwood:
I get this:

```
==> '. /Users/william/ros2/build/ros1_bridge/cmake__build.sh && /usr/bin/make -j4 -l4' in '/Users/william/ros2/build/ros1_bridge'
Traceback (most recent call last):
  File "/Users/william/ros2/src/ros1_bridge/ros1_bridge/__init__.py", line 18, in <module>
    import catkin_pkg
ImportError: No module named 'catkin_pkg'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bin/ros1_bridge_generate_factories", line 11, in <module>
    from ros1_bridge import generate_cpp
  File "/Users/william/ros2/src/ros1_bridge/ros1_bridge/__init__.py", line 25, in <module>
    [python_executable, '-c', 'import catkin_pkg; print(catkin_pkg.__file__)'])
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/subprocess.py", line 607, in check_output
    with Popen(*popenargs, stdout=PIPE, **kwargs) as process:
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/subprocess.py", line 1457, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'python2'
make[2]: *** [CMakeFiles/generate_factories] Error 1
make[1]: *** [CMakeFiles/generate_factories.dir/all] Error 2
make: *** [all] Error 2
make: INTERNAL: Exiting with 5 jobserver tokens available; should be 4!
```

I'm working on a patch.

wjwwood:
I fixed the above with 949ed1e, and after that it works for me on OS X with Indigo.

PR #5, opened 2015-07-30
use pkg-config to find ROS 1 packages

dirk-thomas:
As a follow up of this ticket should I prepare PRs to rename all message packages again to their original names?


esteve:
+1

tfoote:
+1

PR #10, opened 2015-08-28
remove references on smart pointers

dirk-thomas:
Connects to ros2/ros2#94


esteve:
+1

src/listener.cpp
```diff
@@ -19,7 +19,7 @@
 #include <std_msgs/msg/string.hpp>
 
 
-void chatterCallback(const std_msgs::msg::String::ConstSharedPtr & msg)
+void chatterCallback(const std_msgs::msg::String::ConstSharedPtr msg)
```
wjwwood: I think you also need to use `SharedPtr` rather than `ConstSharedPtr` for now. Later we have to figure out how to handle all of these different callback types in a portable way.

dirk-thomas: This PR only removes the reference. I would rather not change the smart pointer type here.

tfoote:
+1

wjwwood:
@dirk-thomas I added a commit which lets it build on my machine. I know this pr was not originally intended to fix build failures, but the "stylistic" changes you started with are actually required now as part of the fix, so I tacked my other changes onto this pr.

dirk-thomas:
lgtm.

wjwwood:
Ok, I added a todo to change this once we support bind again.

PR #11, opened 2015-09-22
support custom mapping rules

dirk-thomas:
See `doc/index.rst` for what it does.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,0 +1,87 @@
+ros1_bridge
+===========
+
+The bridge is a "dual homed" program which acts as a ROS 1 node as well as a ROS 2 node at the same time.
+It can therefore subscribe to messages in one ROS version and publish them into the other ROS version.
+
+The specific messages it is able to convert between ROS 1 and ROS 2 are determined at compile time.
+Automatic mapping rules are being applied based on the names.
+Additionally custom mapping rules can be provided.
+
+
+How are ROS 1 and 2 messages associated with each other?
+--------------------------------------------------------
+
+Automatic mapping between ROS 1 and 2 messages is performed based on the names.
+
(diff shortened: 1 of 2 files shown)


doc/index.rst
```diff
@@ -0,0 +1,87 @@
...
+
+    <export>
+      <ros1_bridge mapping_rules="my_mapping_rules.yaml"/>
+    </export>
+
+The yaml file must also be installed in the ``CMakeLists.txt``::
+
+    install(
+      FILES my_mapping_rules.yaml
+      DESTINATION share/${PROJECT_NAME})
```
wjwwood: Should it also be put in the resource index? Or is that not required?

dirk-thomas: It's currently not using the resource index - simply because it was easier without. It could do so in the future though.

ros1_bridge/__init__.py
```diff
@@ -117,6 +121,20 @@ def get_ros1_messages(rospack=None):
...
     msgs = []
+    rules = []
+    rules.append(MappingRule({
+        'ros1_package_name': 'nav_msgs',
+        'ros1_message_name': 'GridCells',
+        'ros2_package_name': 'nav_msgs',
+        'ros2_message_name': 'GridCells',
+        'fields_1_to_2': {
+            'header': 'header',
+            'cell_width': 'cell_width_new',
```
wjwwood: Is this `_new` suffix intentional? This doesn't look like an automated test to me.

wjwwood:
lgtm, other than comments, but I didn't test it locally (my ROS 1 installation is in the process of being rebuilt).

ros1_bridge/__init__.py
dirk-thomas: Ups, this was a test rule I used. I will remove it from the PR.

PR #12, opened 2015-09-28
use sensor profile since messages might be large which is not implemented in all rmw impl for reliable qos

dirk-thomas:
Required for the presentation today.


dirk-thomas:
Please review.

tfoote:
+1

PR #14, opened 2015-10-29
Update apis

esteve:
This updates the bridge to use the `PublisherBase` / templated `Publisher` and the new version of `create_subscription`

### DIFF
--- a/include/ros1_bridge/bridge.hpp
+++ b/include/ros1_bridge/bridge.hpp
@@ -31,7 +31,7 @@ namespace ros1_bridge
 struct Bridge1to2Handles
 {
   ros::Subscriber ros1_subscriber;
-  rclcpp::publisher::Publisher::SharedPtr ros2_publisher;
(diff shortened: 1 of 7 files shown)


dirk-thomas:
LGTM

dirk-thomas:
No squashing of (at least) commit 1 and 3?

esteve:
It's already merged. I'll keep it in mind next time, but we agreed to squash if we felt like, though.

dirk-thomas:
Well we decided on something more concrete: `Before merging a pull request all changes should be squashed into a small number semantic commits to keep the history clear.` (https://github.com/ros2/ros2/wiki/Developer-Guide#pull-requests).

esteve:
Sounds good, but note the wording: should, not must, and it's at the discretion of the author of the PR. In any case, there's no point in pointing out that the changes weren't squashed right after merging.

PR #15, opened 2015-11-03
fix qos profiles, must use ros1_bridge specific listener

dirk-thomas:


wjwwood:
lgtm, but why does it need the sensor qos profile (presumably for unreliable, but why is that needed)?

dirk-thomas:
Because a reliable (default qos) subscriber won't get any data from a best effort publisher.

README.md
```diff
@@ -112,11 +112,12 @@ The ROS 1 node will start printing the published messages to the console.
 ---
 
 Now we start the ROS 2 listener.
+It is important to use the listener from the `ros1_bridge` package since it uses different QoS parameters then the "normal" `listener`.
```
esteve: then -> than

dirk-thomas: Fixed.

esteve:
+1

PR #16, opened 2015-11-03
ignore unused-parameter warning in ROS 1 message headers

dirk-thomas:
Before http://ci.ros2.org/view/packaging/job/ros2_packaging_osx/17/
After http://ci.ros2.org/view/packaging/job/ros2_packaging_osx/19/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,12 @@ function(build_executable executable)
   endforeach()
 endfunction()
 
+if(NOT WIN32)
+  # ignore warning in ROS 1 message headers
+  # TODO(dirk-thomas) should only be applied to the generated_factories.cpp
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -111,6 +117,9 @@ build_executable(dynamic_bridge src/dynamic_bridge.cpp ROS1_DEPENDENCIES
   DEPENDENCIES generate_factories
   TARGET_DEPENDENCIES ${message_packages})
 
+if(NOT WIN32)
+endif()
+
```
wjwwood: Should this be dropped?

wjwwood:
lgtm other than the comment.

CMakeLists.txt
wjwwood: I went ahead an dropped it.

dirk-thomas: Thanks.

PR #19, opened 2015-11-25
fix build of bridge for namespace change

gerkey:
Packaging jobs (because the CI jobs don't build the bridge):

* http://ci.ros2.org/view/packaging/job/packaging_osx/25
* http://ci.ros2.org/view/packaging/job/packaging_linux/49

Windows doesn't build the bridge, so this change has no effect there.


dirk-thomas:
+1

PR #21, opened 2016-02-17
update from source build instructions

wjwwood:
Some things are no longer necessary.


dirk-thomas:
Should the instructions mention that a current version is needed?

Besides the comment +1.

wjwwood:
I added a note about the required ROS 1 version of `ros_comm` and specifically `rosbag`.

PR #23, opened 2016-04-05
workaround for Python search order

dirk-thomas:
This fixes the packaging jobs which were broken since https://github.com/ros2/rclpy/pull/5#issuecomment-205508263

Before:
* http://ci.ros2.org/view/packaging/job/packaging_linux/199/
* http://ci.ros2.org/view/packaging/job/packaging_osx/177/

After:
* http://ci.ros2.org/view/packaging/job/packaging_linux/200/
* http://ci.ros2.org/view/packaging/job/packaging_osx/178/

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -52,6 +52,18 @@ except ImportError:
         raise
 
 # more ROS 1 imports
+# since ROS 2 should be sourced after ROS 1 it will overlay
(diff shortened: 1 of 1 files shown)


jacquelinekay:
I'm okay with this workaround. Better than changing the import signature. +1

wjwwood:
I agree, +1. Thanks for looking into @dirk-thomas.

Issue #22, opened 2016-03-30
[dynamic_bridge] could not get topic names and types: not implemented


Karsten1987:
I am following [this](https://github.com/ros2/ros1_bridge/blob/master/README.md) tutorial for launching the ros1 bridge. It terminates with a RMW error:
``` 
~/ros2_ws$ dynamic_bridge
terminate called after throwing an instance of 'std::runtime_error'
  what():  could not get topic names and types: not implemented, at /home/karsten/ros2_ws/src/eProsima/ROS-RMW-Fast-RTPS-cpp/rmw_fastrtps_cpp/src/functions.cpp:1775
Aborted (core dumped)
```

Am I doing anything wrong?

jacquelinekay:
The error is not your fault, it's because rmw_fastrtps_cpp doesn't have a complete implementation of the rmw standard yet.

If you have compiled ROS 2 with another DDS implementation, e.g. Opensplice or Connext, you can run the bridge with a different middleware, e.g. `dynamic_bridge__rmw_opensplice_cpp`, and it should work.

If that executable is not available then make sure you've followed the instructions to install Opensplice: https://github.com/ros2/ros2/wiki/Linux-Development-Setup#install-one-or-more-dds-implementations and try rebuilding your workspace.

dirk-thomas:
Closing as duplicate of eProsima/ROS-RMW-Fast-RTPS-cpp#14.

PR #24, opened 2016-04-05
Add note to README about not using fastrtps

jacquelinekay:


gerkey:
Can we point to an existing (FastRTPS) ticket that is tracking the what's missing for the bridge?

README.md
```diff
@@ -25,6 +25,8 @@ The following ROS 1 packages are required to build and use the bridge:
 * `std_msgs`
 * as well as the Python package `rospkg`
 
+**WARNING**: The ROS 1 bridge currently does not work if ROS 2 is built with the FastRTPS middleware.
+If you installed ROS 2 from source, make sure you installed it with another supported middleware (Opensplice or Connext).
```
dirk-thomas: If you installed ROS 2 from source, make sure you **built** it with ... ?

jacquelinekay:
https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/14 is the original ticket for the missing rmw functions that are called by the rclcpp functions used by the bridge to collect existing topic names. There are references to other issues there (e.g. #22 was specific to the ROS 1 bridge but was closed as a duplicate of the first issue)

jacquelinekay:
Addressed comments.

gerkey:
+1

kartikmohta:
I think this works now since the functions have been implemented in Fast RTPS (https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/56), so the note can be removed from the README

dirk-thomas:
@kartikmohta Thank you for pointing this out. I have reverted the commit in bd55d29aab567607337b3de9b4685e0ec10536de.

PR #27, opened 2016-07-17
fix link order

### LINKED ISSUE #26: Error on start of ros1_bridge_listener
terminate called after throwing an instance of `std::runtime_error`:

```
what():  could not create subscription: type support implementation 'introspection_cpp'(0x7f59fceae669) does not match rmw implementation 'opensplice_static'(0x7f59fb535659), at /home/rohan/ros2_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_subscription.cpp:66, at /home/rohan/ros2_ws/src/ros2/rcl/rcl/src/rcl/subscription.c:76
Aborted (core dumped)
```

Running on xenial with ROS kinetic, and ROS2 alpha7.
I compiled ROS2 from source, and disabled fastrtps, so I only have opensplice.

dirk-thomas:
Fixes #26. Connect to #26.

Please review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,8 +65,8 @@ function(custom_executable target)
   add_executable(${target}${target_suffix}
     ${ARG_UNPARSED_ARGUMENTS})
   ament_target_dependencies(${target}${target_suffix}
-    ${ARG_TARGET_DEPENDENCIES}
-    "rclcpp${target_suffix}")
+    "rclcpp${target_suffix}"
+    ${ARG_TARGET_DEPENDENCIES})
   if(ARG_ROS1_DEPENDENCIES)
     ament_target_dependencies(${target}${target_suffix}
       "ros1_roscpp"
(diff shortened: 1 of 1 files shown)


gerkey:
It's hard to judge the impact of the change, but it makes sense to me. +1

dirk-thomas:
A message package exports the following libraries:
* one for the data structures and
* one for each type support

When this package is being linked first any of the type support packages satisfies the few required symbols of the executable / library using the messages (since no type support specific symbols are being used anywhere). So the linker will pick one of them (for whatever reason, e.g. order) which might be "wrong".

rclcpp is being linked against a specific rmw implementation. That rmw implementation has the need for additional type support symbols. So when rclcpp is being linked first and the message package is linked afterwards only one of the type support libraries satisfies these symbols and therefore the linker can pick the "right" one.

gerkey:
Thanks for the helpful explanation. Does it follow that al packages using both rclcpp and messages must always link rclcpp before the messages?

dirk-thomas:
Yes, all target linking against both require that order.

Issue #32, opened 2016-10-17
publishers should use reliable


dirk-thomas:
Currently all publishers created by the bridge are best effort and therefore reliable subscribers are unable to receive the bridged data directly.

PR #37, opened 2016-12-04
Change ROS 2 publishers to reliable so any reliability profile can connect

dhood:
fixes https://github.com/ros2/ros1_bridge/issues/32
connects to ros2/ros1_bridge#32

note that they were best effort because of large messages https://github.com/ros2/ros1_bridge/pull/12

jenkins nodes are currently having issues building the bridge but I've tested both executables locally


include/ros1_bridge/factory.hpp
```diff
@@ -45,7 +45,7 @@ class Factory : public FactoryInterface
     const std::string & topic_name,
     size_t queue_size)
   {
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
+    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
```
dirk-thomas: Since this uses the default it can just omit the profile argument.

Same below.

README.md
```diff
@@ -115,12 +115,11 @@ The ROS 1 node will start printing the published messages to the console.
 ---
 
 Now we start the ROS 2 listener.
-It is important to use the listener from the `ros1_bridge` package since it uses different QoS parameters than the "normal" `listener`.
 
 ```
 # Shell D:
 . <install-space-with-ros2>/setup.bash
-ros1_bridge_listener
+listener
```
dirk-thomas: Maybe it would be helpful to mention which package provides the `listener`.

include/ros1_bridge/factory.hpp
dhood: I didn't do it here because it's used in the next line but I did for `simple_bridge_1_to_2` in https://github.com/ros2/ros1_bridge/pull/37/commits/08881e49b207237b8372539fddaa84df9e77f8e6

README.md
dhood: https://github.com/ros2/ros1_bridge/pull/37/commits/08881e49b207237b8372539fddaa84df9e77f8e6

although we do now rely on users not installing another executable with the same name, but that will probably be true for most demos

mikaelarguedas: should `rclcpp_examples` be listed as a dependency then ?

dhood: it's just a dependency for the tutorial, so I don't think so

mikaelarguedas: ok sounds good, this executable is now in the `rclcpp_tutorials` package (for the readme)

dhood:
could I get a review on this please

mikaelarguedas:
lgtm

PR #39, opened 2016-12-04
Update readme for xenial+kinetic

dhood:
e.g. `rqt` can have the topics specified through command line now, so 'shell D' is not necessary in the image demo (I left the naming of shell E the same because of the screenshot)

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ rqt_image_view /image
 
 Now we start the ROS 2 image publisher:
 ```
-# Shell E:
+# Shell D:
 . <workspace-with-ros2>/install/setup.bash
 cam2image
 ```
@@ -257,7 +257,7 @@ By publishing either `true` or `false` to the `flip_image` topic, the camera nod
 You can either use the `Message Publisher` plugin in `rqt` to publish a `std_msgs/Bool` message on the topic `flip_image`, or run one of the two following `rostopic` commands:
 
 ```
-# Shell F:
+# Shell E:
 . /opt/ros/kinetic/setup.bash
 # Or, on OSX, something like:
 # . ~/ros_catkin_ws/install_isolated/setup.bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,6 +66,8 @@ Then build just the ROS 1 bridge with `-j1`:
 src/ament/ament_tools/scripts/ament.py build --build-tests --symlink-install -j1 --only ros1_bridge
 ```
 
+*Note:* If you have built your ROS 2 workspace using `--isolated`, you will need to add the message/service packages you want to bridge to your path before building the bridge, e.g. by sourcing `local_setup.*`.
```
dhood: the packaging job (non-isolated) doesn't seem to do it. I understood that in non-isolated builds the packages installed are on the path by default?

dirk-thomas: No, in both cases a file needs to be sourced.

README.md
```diff
@@ -2,8 +2,10 @@
 
 This package provides a network bridge which enables to exchange messages between ROS 1 and ROS 2.
 
-The bridge is currently implemented in C++ simply because there is no Python API for ROS 2 yet.
-Therefore it is limited to all the message types available at compile time.
+The bridge has currently only been implemented in C++ as at the time the Python API for ROS 2 had not been developed.
+Because of this its support is limited to only the message/service types available at compile time of the bridge.
+The bridge included with the prebuilt ROS 2 binaries supports the interface packages listed in the [ros2/common_interfaces repository.](https://github.com/ros2/common_interfaces)
```
dhood: since `rcl_intefaces` don't have mappings associated with them they are not included in the packaged bridge. I agree that the statement should be made more general: https://github.com/ros2/ros1_bridge/commit/962c85bf369c45afdbaab8b5d7f4cc6ac0317ebe#diff-04c6e90faac2675aa89e2176d2eec7d8R7

PR #41, opened 2016-12-05
update shell title in image

dirk-thomas:
Connect to #39.


mikaelarguedas:
Should the image reflect that we advise people to use ROS kinetic rather than indigo ?

dhood:
thanks for the image @dirk-thomas 
@mikaelarguedas I will add a comment that the image is using indigo still

dhood:
yep I'll merge this when #39 is good

PR #42, opened 2016-12-06
Add option for bridging all topics

dhood:
without this, getting a topic to be bridged requires starting up temporary subscribers on the other side. introspection tools e.g. `rostopic list` can't see ROS 2 publishers, and tools such as `rosbag record --all` aren't very usable.

With this, bridges will be made for any topic that has a known mapping, whether subscribers are on the other side or not.

right now it is trying to make a bridge for a topic even if it doesn't have a known mapping, so you get e.g. `failed to create 1to2 bridge for topic 'rosout_agg' with ROS 1 type 'rosgraph_msgs/Log' and ROS 2 type '': No template specialization for the pair` printed out. I will fix it by adding a function that check if a factory exists for the type before forcibly creating the mapping.

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -72,9 +72,14 @@ def generate_cpp(output_path, template_dir):
     data.update(generate_services())
     unique_package_names = set(data["ros2_package_names_msg"] + data["ros2_package_names_srv"])
     data["ros2_package_names"] = list(unique_package_names)
+
     template_file = os.path.join(template_dir, 'get_factory.cpp.em')
     output_file = os.path.join(output_path, 'get_factory.cpp')
     expand_template(template_file, data, output_file)
+
+    template_file = os.path.join(template_dir, 'get_mappings.cpp.em')
+    expand_template(template_file, data, os.path.join(output_path, 'get_mappings.cpp'))
+
     for ros2_package_name in data["ros2_package_names"]:
         for extension in ['cpp', 'hpp']:
             data_pkg = {
@@ -150,6 +155,7 @@ def generate_messages():
         'mappings': ordered_mappings,
         'ros2_package_names_msg': ros2_package_names
     }
+
 
 def generate_services():
     ros1_srvs = get_ros1_services()
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -69,6 +69,12 @@ import rosmsg
 
 def generate_cpp(output_path, template_dir):
     data = generate_messages()
+
+    template_file = os.path.join(template_dir, 'get_mappings.cpp.em')
+    output_file = os.path.join(output_path, 'get_mappings.cpp')
+    data_for_template = {'mappings': data['mappings']}
+    expand_template(template_file, data_for_template, output_file)
+
     data.update(generate_services())
     unique_package_names = set(data["ros2_package_names_msg"] + data["ros2_package_names_srv"])
     data["ros2_package_names"] = list(unique_package_names)
@@ -76,9 +82,6 @@ def generate_cpp(output_path, template_dir):
     template_file = os.path.join(template_dir, 'get_factory.cpp.em')
     output_file = os.path.join(output_path, 'get_factory.cpp')
     expand_template(template_file, data, output_file)
-
-    template_file = os.path.join(template_dir, 'get_mappings.cpp.em')
-    expand_template(template_file, data, os.path.join(output_path, 'get_mappings.cpp'))
 
     for ros2_package_name in data["ros2_package_names"]:
         for extension in ['cpp', 'hpp']:
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -129,7 +129,7 @@ void update_bridge(
         continue;
       }
       // update the ROS 2 type name to be that of the anticipated bridged type
-      //TODO(dhood): support non 1-1 "bridge-all" mappings
+      // TODO(dhood): support non 1-1 "bridge-all" mappings
       bool mapping_found = ros1_bridge::get_1to2_mapping(ros1_type_name, ros2_type_name);
       if (!mapping_found) {
         // printf("No known mapping for ROS 1 type '%s'\n", ros1_type_name.c_str());
@@ -190,9 +190,9 @@ void update_bridge(
         continue;
       }
       // update the ROS 1 type name to be that of the anticipated bridged type
+      // TODO(dhood): support non 1-1 "bridge-all" mappings
       bool mapping_found = ros1_bridge::get_2to1_mapping(ros2_type_name, ros1_type_name);
-      if (!mapping_found)
-      {
+      if (!mapping_found) {
         // printf("No known mapping for ROS 2 type '%s'\n", ros2_type_name.c_str());
         continue;
       }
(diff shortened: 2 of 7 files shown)


dirk-thomas:
This approach could become problematic if we ever have something beyond a 1:1 mapping. If a single type in one ROS version can be mapped to more than one type in the other version it is unknown which mapping should be created.

dhood:
Thanks for the point about the non-1:1 mappings. it doesn't seem like it will be much effort to make [this new method](https://github.com/ros2/ros1_bridge/pull/42/commits/8485326b578cc65c77380f75cf59f7fbaa05ed17#diff-ee525d02c311b44f9fefa22ad669d365R50) return all possible mappings and forcing them all - what do you think?

Also, what do you think is best for this signature https://github.com/ros2/ros1_bridge/pull/42/commits/8485326b578cc65c77380f75cf59f7fbaa05ed17#diff-ee525d02c311b44f9fefa22ad669d365R56 : how it is now, where outputs are listed after inputs, or switching them so that `ros1_type_name` comes before `ros2_type_name` (for consistency with `get_1to2_mapping`)

dhood:
alternatively we could not support non-1:1 mappings at this point. I think we will already run into issues on master if you have multiple mappings from ROS 2 to ROS 1 because the ROS 1 subscribers all get assigned type `""`

dirk-thomas:
> Thanks for the point about the non-1:1 mappings. it doesn't seem like it will be much effort to make [this new method](https://github.com/ros2/ros1_bridge/pull/42/commits/8485326b578cc65c77380f75cf59f7fbaa05ed17#diff-ee525d02c311b44f9fefa22ad669d365R50) return all possible mappings and forcing them all - what do you think?

That sounds good.

> Also, what do you think is best for this signature https://github.com/ros2/ros1_bridge/pull/42/commits/8485326b578cc65c77380f75cf59f7fbaa05ed17#diff-ee525d02c311b44f9fefa22ad669d365R56 : how it is now, where outputs are listed after inputs, or switching them so that `ros1_type_name` comes before `ros2_type_name` (for consistency with `get_1to2_mapping`)

For a function calles `get_XtoY_mapping` I would expect the arguments to be in the order as mentioned in the name, so first `X` and then `Y`. The fact that there is also a function `getYtoX_mapping` doesn't matter imo.

dhood:
I've rebased this after https://github.com/ros2/ros1_bridge/pull/44
I haven't accommodated non-1:1 mappings but think that this could still be reviewed as-is

src/dynamic_bridge.cpp
```diff
@@ -51,6 +51,50 @@ struct Bridge2to1HandlesAndMessageTypes
...
+  return std::find(args.begin(), args.end(), option) != args.end();
+}
+
+bool get_flag_option(const std::vector<std::string> & args, const std::string & option)
+{
+  auto it = std::find(args.begin(), args.end(), option);
+  if (it != args.end()) {
+    return true;
+  }
+  return false;
```
dirk-thomas: These lines can be written as:

```return it != args.end()```

File: src/dynamic_bridge.cpp [hunk 2]
```diff
@@ -60,23 +104,36 @@ void update_bridge(
...
     // identify topics available as ROS 1 publishers as well as ROS 2 subscribers
     auto topic_name = ros1_publisher.first;
+    std::string ros1_type_name = ros1_publisher.second;
+    std::string ros2_type_name;
+
     auto ros2_subscriber = ros2_subscribers.find(topic_name);
     if (ros2_subscriber == ros2_subscribers.end()) {
-      continue;
+      if (bridge_all_1to2_topics) {
+        // update the ROS 2 type name to be the anticipate the bridged type
```
dirk-thomas: Missing word?

src/dynamic_bridge.cpp
```diff
@@ -60,23 +104,36 @@ void update_bridge(
...
+        bool mapping_found = ros1_bridge::get_1to2_mapping(ros1_type_name, ros2_type_name);
+        if (!mapping_found) {
+          // printf("No known mapping for ROS 1 type '%s'\n", ros1_type_name.c_str());
+          continue;
+        }
+        // printf("topic name '%s' has ROS 2 publishers\n", topic_name.c_str());
+      } else {
+        continue;
+      }
```
dirk-thomas: The nesting is pretty deep with late `else` blocks and `continue`. I would suggest inverting the logic to `continue` early:

```
if (!bridge_all_1to2_topics) {
  continue;
}
bool mapping_found = ros1_bridge::get_1to2_mapping(ros1_type_name, ros2_type_name);
if (!mapping_found) {
  continue;
}
```

src/dynamic_bridge.cpp
```diff
@@ -114,18 +171,31 @@ void update_bridge(
...
+    std::string ros2_type_name = ros2_publisher.second;
+    std::string ros1_type_name;
+
+    auto ros1_subscriber = ros1_subscribers.find(topic_name);
+    if (ros1_subscriber == ros1_subscribers.end()) {
+      if (bridge_all_2to1_topics) {
+        // update the ROS 1 type name to be the anticipate the bridged type
+        bool mapping_found = ros1_bridge::get_2to1_mapping(ros2_type_name, ros1_type_name);
+        if (!mapping_found)
+        {
```
dirk-thomas: Same here. And the `{` should be trailing the previous line.

src/dynamic_bridge.cpp
```diff
@@ -200,6 +270,15 @@ void update_bridge(
 
 int main(int argc, char * argv[])
 {
+  bool output_topic_introspection;
+  bool bridge_all_1to2_topics;
+  bool bridge_all_2to1_topics;
+  if (!parse_command_options(argc, argv,
+    output_topic_introspection, bridge_all_1to2_topics, bridge_all_2to1_topics))
```
dirk-thomas: Please wrap before `argc` ("either all or none but not some").

ros1_bridge/__init__.py
```diff
@@ -114,6 +114,10 @@ def generate_cpp(output_path, template_dir):
     template_file = os.path.join(template_dir, 'get_factory.cpp.em')
     expand_template(template_file, data, os.path.join(output_path, 'get_factory.cpp'))
 
+    data = {'mappings': ordered_mappings}
+    template_file = os.path.join(template_dir, 'get_mappings.cpp.em')
```
dirk-thomas: Where is this template?

dhood: oops - added in https://github.com/ros2/ros1_bridge/pull/42/commits/e0c68e870878896533b34c7e5396a24a2f1f1d5e

dhood:
Merge conflicts addressed after services were added.
I am not sure why https://github.com/ros2/ros1_bridge/pull/42/commits/4d26a312210a3e381806f08e03dc75e81f380867 was necessary to avoid empy `em.Error: interpreter globals collision` errors but I can confirm that bridging all topics is working for the packaged bridge.

Here's a job with the `bridge-all*` options: http://ci.ros2.org/view/packaging/job/packaging_linux-ci/1/

dirk-thomas:
You can't reuse the same dictionary for multiple empy evaluations. You should pass a "new" dict to each evaluate call instead.

dhood:
ah, thank you :smile:
in that case I propose we merge this.
When testing this I did notice that the service bridging wasn't working (at least now how I expected it to). I ran a job from master and confirmed it wasn't only on this branch. (I mention it in case you're testing the packaging job and wondering the same)

dirk-thomas:
Sounds good to me.

PR #44, opened 2016-12-06
Separate compilation units

### LINKED ISSUE #40: split generated code in separate compilation unit
Currently all generated template specializations are generated in a single `.cpp` file which requires *a lot* of memory to compile.

dirk-thomas:
Fixes #40. Connect to #40.

I propose to merge this PR first. I am happy to look into updating the existing PRs since this one moves quite some code around.


dhood:
what would be the recommended build invocation now? the same as before but without the `-j1`?

dirk-thomas:
Yes, same as before. To use `j1` or not still depends on the available resources. The build will now use much less resources per compilation unit. But if you run it with potentially eight cores it might still need more then the system has.

dhood:
ok so without modification to the packaging job build invocations, do I understand correctly that this should allow the [failing linux packaging job](http://ci.ros2.org/view/packaging/job/packaging_linux/500/console) to run again without having the RAM increased? (I have launched http://ci.ros2.org/view/packaging/job/packaging_linux/501/ to check)

mikaelarguedas:
AFAI understand it should actually be decreased compare to the previous
runs without the messages we enabled recently

dirk-thomas:
Yes, the packaging job passes with this patch.

resource/pkg_factories.cpp.em
```diff
@@ -1,53 +1,32 @@
...
 
+// include builtin interfaces
+#include <ros1_bridge/convert_builtin_interfaces.hpp>
 
 namespace ros1_bridge
 {
 
 std::shared_ptr<FactoryInterface>
-get_factory(const std::string & ros1_type_name, const std::string & ros2_type_name)
+get_factory_@(ros2_package_name)(const std::string & ros1_type_name, const std::string & ros2_type_name)
```
dhood: I get the feeling that this might also one day suffer from the non-1:1 mapping situation mentioned [here](https://github.com/ros2/ros1_bridge/pull/42#issuecomment-265220872) (I am not 100% sure). either way I don't think it should block it.

dirk-thomas: The non-1:1 problem doesn't apply here. This patch only groups all specialization by their ROS 2 package name. Every mapping always has exactly one ROS 2 package associated with it. If there are multiple mapping for the same type within the package that is being handled within the generated conditions which compare the passed string values.

dhood: thank you for the explanation!

PR #45, opened 2016-12-08
support for bridging services

dirk-thomas:
This will replace #36. I have tried to keep the history of the previous PR (without the tests yet). Based on that I have resolved the merge conflicts (due to #44) and made it compile again. I haven't yet spend the time to test the bridge or refactor the code to avoid various duplication (search for `TODO(dirk-thomas)`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ The following ROS 1 packages are required to build and use the bridge:
 * `std_msgs`
 * as well as the Python package `rospkg`
 
+**Note:** Bridge will connect only services with the same package name and service name so make sure that they are available when you build ros1_bridge.
 
 ### Build the bridge from source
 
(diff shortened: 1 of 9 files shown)


dhood:
@dirk-thomas the second commit in #36 wasn't just adding tests, it was also addressing comments from a review of yours. as a result, this new PR doesn't seem to fully capture all of the work that was in #36, as https://github.com/ros2/ros1_bridge/pull/36/commits/84f02e72ad6c92ba10aaaf9d692ed92cd36fd79a had already addressed some of the TODOs you are referring to

I am not saying that you should rebase the second commit, but I wanted to point out in case it was something you weren't aware of

dirk-thomas:
Good point. I will try to manually merge them when I get back to this.

dhood:
ok. we can then add the "connects to ros2/examples#136" tag too

haueck:
I added one more commit to the https://github.com/ros2/ros1_bridge/pull/36. I can rebase my fork with ros2/ros1_bridge/master and it will be possible to merge my pull request without any conflicts. Maybe it will be easier that way, what do you think?

PR #48, opened 2016-12-14
Rework after the review (#36)

haueck:
* Removed separate file for mapping services
* Other minor changes


dirk-thomas:
Thank you for the fast update.

PR #49, opened 2016-12-14
cleanup and pass all linters

dirk-thomas:
This only cleans up the code and makes all linters pass. The functional tests still fail. Running the dynamic bridge manually works. I will look into them separately.


test/test_ros2_client.cpp
```diff
@@ -12,24 +12,26 @@
...
+  auto request = std::make_shared<diagnostic_msgs::srv::SelfTest::Request>();
 
   if (!client->wait_for_service(4_s)) {
     throw std::runtime_error("Service is not available");
   }
 
   auto future = client->async_send_request(request);
-  if (rclcpp::spin_until_future_complete(node, future, 2_s) == rclcpp::executor::FutureReturnCode::SUCCESS)
+  if (
+    rclcpp::spin_until_future_complete(node, future, 2_s) == \
```
dirk-thomas: No, its not. Removed.

PR #57, opened 2017-02-16
correctly export libraries and fix LD_LIB path

Karsten1987:


wjwwood:
What is `LD_LIB`?

Karsten1987:
That's me being lazy to type LD_LIBRARY_PATH :-)

PR #58, opened 2017-02-17
Fixed crashing discovery of ROS2 services

haueck:


wjwwood:
Unfortunately our CI doesn't currently test the ROS 1 bridge, but I can run a Linux packaging job with this change to check for compiler errors and warnings:

[Build Status](http://ci.ros2.org/job/packaging_linux/585/)

@Karsten1987 or @jhoare could either of you manually test this patch?

Karsten1987:
that looks good to!
+1

wjwwood:
Thanks @haueck!

PR #68, opened 2017-05-30
fix race conditions

### LINKED ISSUE #61: crash in map lookup
I get the following crash in dynamic_bridge when I shut down my sender:
```
failed to create 2to1 bridge for topic 'map' with ROS 2 type '����' and ROS 1 type 'nav_msgs/OccupancyGrid': No template specialization for the pair
check the list of supported pairs with the `--print-pairs` option
created 2to1 bridge for topic 'map' with ROS 2 type 'nav_msgs/OccupancyGrid' and ROS 1 type 'nav_msgs/OccupancyGrid'
```
OccupancyGrid works fine for some period of time before the failure. The memory for the string used as the map key is somehow getting trampled. Hence the weird characters in the type name. Note: I've been using the opensplice middleware (so that could be the trampler).

dirk-thomas:
Before the polling lambda modified the maps while the `update_bridge` is iterating over them.

Now these changes are build up in a separate variable before the variables passed by reference to `update_bridge`. When they are swapped the global mutex is being locked.

Fixes #61.

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -554,8 +554,8 @@ int main(int argc, char * argv[])
         return;
       }
 
-      ros1_publishers.clear();
-      ros1_subscribers.clear();
+      std::map<std::string, std::string> current_ros1_publishers;
+      std::map<std::string, std::string> current_ros1_subscribers;
       for (auto topic : topics) {
         auto topic_name = topic.name;
         if (topic_name.compare(0, 1, "/") == 0) {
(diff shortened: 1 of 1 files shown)


src/dynamic_bridge.cpp
```diff
@@ -554,8 +554,8 @@ int main(int argc, char * argv[])
         return;
       }
 
-      ros1_publishers.clear();
-      ros1_subscribers.clear();
+      std::map<std::string, std::string> current_ros1_publishers;
+      std::map<std::string, std::string> current_ros1_subscribers;
```
mikaelarguedas: should we use "active_ros1_publishers"/"active_ros1_subscribers" to match the naming of the equivalent "active_ros1_services" variable? In general the naming doesn't match between publishers/subscribers and services variables but I think we can at least do this bit in this PR.

deng02:
I saw this bug during testing bridging ros1 to ros2 topics; confirm that this change fixed it for us.

src/dynamic_bridge.cpp
dirk-thomas: This PR only fixes the race condition. `active` is imo not a good prefix. Therefore I chose `current`. But I would rather not touch any unrelated code. Any kind of renaming / unification can be done in a separate PR.

dirk-thomas:
> a Linux packaging job build.

@wjwwood I was not planning to run a CI build since the tests won't be able to show any improvement but only proof that the code still compiles. Anyway this is a build of this branch using a temporary job: [Build Status](http://ci.ros2.org/job/test_packaging_linux/1/)

PR #69, opened 2017-05-31
add parameter-based bridge

dirk-thomas:
This variation of the bridge allows the user to configure a list of topics with their types and queue sizes on the ROS 1 parameter server and bridge exactly that set of topics.

[Build Status](http://ci.ros2.org/job/ci_packaging_linux/10/)

### DIFF
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -1,0 +1,102 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <list>
+#include <string>
+
+// include ROS 1
+#ifdef __clang__
+# pragma clang diagnostic push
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Waiting for review.

dirk-thomas:
Waiting for review.

src/parameter_bridge.cpp
```diff
@@ -0,0 +1,96 @@
...
+    parameter_name = argv[1];
+  }
+
+  XmlRpc::XmlRpcValue topics;
+  if (
+    ros1_node.getParam(parameter_name, topics) &&
+    topics.getType() == XmlRpc::XmlRpcValue::TypeArray)
+  {
+    for (size_t i = 0; i < static_cast<size_t>(topics.size()); ++i) {
+      std::string topic_name = static_cast<std::string>(topics[i]["topic"]);
```
dhood: There's a check that the topics param is an array, but not that the elements of the array are dictionaries. this will crash with an XmlRpcException if they're not.

dirk-thomas:
This was developed together with ClearPath to support their use case of trying the bridge. I would rather avoid spending any more time on this unless absolutely necessary.

dhood:
sure, we can live without additional error handling, but adding an example of the `topics` parameter is necessary IMO

dirk-thomas:
> sure, we can live without additional error handling, but adding an example of the topics parameter is necessary IMO

Where would you like the example to be added?

dhood:
only the dynamic bridge really has a readme, so even if it's just a comment in the code itself (like how the static bridge is sort of [code documented](https://github.com/ros2/ros1_bridge/blob/a22159447f1247fd41c373cb68224beb05c31cac/src/static_bridge.cpp#L43)) that would work for now

dirk-thomas:
Sounds good: see https://github.com/ros2/ros1_bridge/pull/69/files#diff-34645027425214113f4c87f76275745cR46

PR #67, opened 2017-05-26
Find ROS1 message packages

mikaelarguedas:
Currently the bridge doesnt compile if you have an overlay ROS1 workspace. The messages are found and added to the factories but then fail to compile because we never find_package them.

the first commits rename a variable to make it obvious it's referring to ros2 packages.
the second commit call rosmsg and rossrv to get the list of ROS1 message packages and find_package them
the third one makes the shared lib link against these  packages
commit 3 and 4 are just fixups/ cleanup

PR #71, opened 2017-06-10
Fix topic matching

wjwwood:
This just removes code that was inplace to adapt ros 1 topics and services that have leading `/` to ros 2 topics and services which did not, until recently.

This should fix https://github.com/ros2/build_cop/issues/28.

This should also remove the need to use `--bridge-all-topics`, @Karsten1987 @Kukanani fyi.

---

I also fixed an issue where the dynamic bridge would crash if the header info for a service was incomplete. The case was being caught and a message was being printed to stderr, but later the code would crash because there was a partially filled out entry in the map. So I just remove it when this case happens now. That prevents the crash, but I don't know why the service info is incomplete in these cases.

Packaging job:

[Build Status](http://ci.ros2.org/job/ci_packaging_linux/12/)

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -428,7 +428,8 @@ void get_ros1_service_info(
     std::string value;
     auto success = header_in.getValue(field, value);
     if (!success) {
-      fprintf(stderr, "Failed to read %s from a header\n", field.data());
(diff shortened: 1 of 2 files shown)


Kukanani:
Can confirm that this makes `dynamic_bridge` work without `--bridge-all-2to1-topics`, when also rebased on #67.

PR #72, opened 2017-06-12
Use multimap for get_all_mappings_2to1

dhood:
When custom mappings are being defined, we might have a 1-to-n mapping from ROS 1 to ROS 2. E.g. both tf2_msgs/TFMessage and tf/tfMessage mapping to tf2_msgs/TFMessage. The bridge already supports this use case, but storage of a multimap is needed to make `--print-pairs` show all supported mappings.

With this PR:
```
$ dynamic_bridge --print-pairs | grep tf

  - 'tf2_msgs/TF2Error' (ROS 2) <=> 'tf2_msgs/TF2Error' (ROS 1)

  - 'tf2_msgs/TFMessage' (ROS 2) <=> 'tf2_msgs/TFMessage' (ROS 1)

  - 'tf2_msgs/TFMessage' (ROS 2) <=> 'tf/tfMessage' (ROS 1)
```

CI from last week [Build Status](http://ci.ros2.org/job/ci_packaging_linux/1/)

### DIFF
--- a/resource/get_mappings.cpp.em
+++ b/resource/get_mappings.cpp.em
@@ -59,10 +59,10 @@ get_2to1_mapping(const std::string & ros2_type_name, std::string & ros1_type_nam
   return false;
 }
 
-std::map<std::string, std::string>
+std::multimap<std::string, std::string>
 get_all_mappings_2to1()
 {
-  static std::map<std::string, std::string> mappings = {
+  static std::multimap<std::string, std::string> mappings = {
 @[for m in mappings]@
     {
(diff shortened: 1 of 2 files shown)


dirk-thomas:
The logic in `get_1to2_mapping` and `get_2to1_mapping` won't work if the mapping is not 1-to-1 (since the first `if` condition which matches will choose the mapping even if there is another one below). So I don't think this can be merged as-is.

dhood:
yep, you're right. non 1-to-1 mapping users will have more issues than just this. we can revisit if/when we support that better

PR #75, opened 2017-06-19
install executables in package specific path

dirk-thomas:
Connect to ros2/ros2#364.

Ready for review.


mikaelarguedas:
IMO it's good to do it now rather than the next beta so that people get used to the "final" way of accessing and running executable as early as possible. It's a good point that this will increase the amount of doc and testing for this specific beta but I think we have enough buffer to handle it and will be beneficial in the long run.

Issue #78, opened 2017-06-27
Unexpected AF_INET Problems


allenh1:
Running the bridge with both fastRTPS and connexts leads to the same behavior after a large amount of messages (consistently ~950 for connext, ~1100 for fastRTPS):

```
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
Couldn't find an AF_INET address for [machina.local]
[...]

PR #79, opened 2017-06-28
Fix fd leak

allenh1:
I tested this by doing a `ulimit -n 100` and then running the bridge. It's past 7000 msgs without the "out of file descriptors" message showing up, so I think this solves it. I was kind of surprised that the `ros::TransportTCP` destructor didn't close the socket.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/336/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2288/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2968/)

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -34,6 +34,7 @@
 
 // include ROS 2
 #include "rclcpp/rclcpp.hpp"
+#include "rclcpp/scope_exit.hpp"
 
 #include "ros1_bridge/bridge.hpp"
 
@@ -387,6 +388,7 @@ void get_ros1_service_info(
     return;
   }
   ros::TransportTCPPtr transport(new ros::TransportTCP(nullptr, ros::TransportTCP::SYNCHRONOUS));
+  auto transport_exit = rclcpp::make_scope_exit([&transport](){ transport->close(); });
   if (!transport->connect(host, port)) {
     fprintf(stderr, "Failed to connect to %s:%d\n", host.data(), port);
     return;
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -388,7 +388,9 @@ void get_ros1_service_info(
     return;
   }
   ros::TransportTCPPtr transport(new ros::TransportTCP(nullptr, ros::TransportTCP::SYNCHRONOUS));
-  auto transport_exit = rclcpp::make_scope_exit([&transport](){ transport->close(); });
+  auto transport_exit = rclcpp::make_scope_exit([&transport]() {
+    transport->close();
+  });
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Same missing for the return in line 428. Since that might be fragile and break again in the future can you please make use something like `make_scope_exit` to ensure the transport is always closed when it goes out of scope.

allenh1:
Done. Changed to `make_scope_exit` (which is super nifty and has changed my life forever, by the way). Should I run the CI again?

tfoote:
Yes, that would be good since the code changed notably since the last revision. It should be fine, but that's why we have automated testing. You can cancel the above jobs still in flight to save time.

dhood:
The linux CI "packaging" job is the only one that builds and runs the bridge tests: http://ci.ros2.org/job/ci_packaging_linux/

allenh1:
* Packaging Job [Build Status](http://ci.ros2.org/job/ci_packaging_linux/31/)

allenh1:
@dhood thank you!

dhood:
no worries, I'll test it out once the job finishes to see if it fixes it for me

dhood:
Testing results: I got through playing the whole rosbag that usually fails for me with the dynamic bridge. It was passing rosbag data (-r 5) from ROS 1 to AMCL in ROS 2 and the pose array etc back to rviz in ROS 1.

It gives a lot of `Failed to connect to osrf-esteve:55738` error messages now where I don't think it used to, but the bridge is functional.

By the way, @allenh1, if someone else runs a CI packaging job the job artifact (including the bridge) will be from the most recent job.

src/dynamic_bridge.cpp
```diff
@@ -387,6 +388,9 @@ void get_ros1_service_info(
     return;
   }
   ros::TransportTCPPtr transport(new ros::TransportTCP(nullptr, ros::TransportTCP::SYNCHRONOUS));
+  auto transport_exit = rclcpp::make_scope_exit([&transport]() {
```
dirk-thomas: boost/std shared pointer shouldn't be passed by reference since that undermines the reference counting. Please capture it by value instead which will make sure that the lambda has its own reference which contributes to the count.

allenh1: I was wondering about that, actually. Will do.

allenh1: Fixed.

allenh1:
@dhood That sounds like something related to #78. I'm investigating that one tomorrow (or later today, I suppose). If you have any ideas as to what might be causing this, let me know!

PR #81, opened 2017-07-03
use ros2 run

dirk-thomas:


README.md
```diff
@@ -194,17 +194,17 @@ roscore
...
 . <install-space-with-bridge>/setup.bash
 export ROS_MASTER_URI=http://localhost:11311
-dynamic_bridge
+ros2 run ros1_bridge dynamic_bridge
 ```
 
 ---
 
-Now we start the ROS 2 talker from the `demo_nodes_cpp` ROS 2 package.
+Now we start the ROS 2 talker from the `demo_nodes_py` ROS 2 package.
```
mikaelarguedas: any benefit on switching to the python example here ?

dirk-thomas: It just consistently uses different languages between `talker` and `listener`.

mikaelarguedas: sounds good

Issue #82, opened 2017-07-24
dynamic_bridge: No executable found


volkandre:
After a fresh ros2 beta2 source install I tried `ros2 run ros1_bridge dynamic_bridge`, but got a `No executable found` error.
So I tried to compile ros1_bridge manually, by typing: `src/ament/ament_tools/scripts/ament.py build --build-tests --symlink-install -j1 --only ros1_bridge` (as suggested in the ros1_bridge Readme file). I sourced my ROS1 environment beforehand.
But ament did not create any executables, although there was no error in the output: 

<pre>Process package 'ros1_bridge' with context:
--------------------------------------------------------------------------------
 source_space => /home/username/ros2_ws/src/ros2/ros1_bridge
  build_space => /home/username/ros2_ws/build/ros1_bridge
install_space => /home/username/ros2_ws/install
   make_flags => -j1
  build_tests => True
--------------------------------------------------------------------------------
+++ Building 'ros1_bridge'
==> '. /home/username/ros2_ws/build/ros1_bridge/cmake__build.sh && /usr/bin/make cmake_check_build_system' in '/home/username/ros2_ws/build/ros1_bridge'
==> '. /home/username/ros2_ws/build/ros1_bridge/cmake__build.sh && /usr/bin/make -j1' in '/home/username/ros2_ws/build/ros1_bridge'
+++ Installing 'ros1_bridge'
==> '. /home/username/ros2_ws/build/ros1_bridge/cmake__install.sh && /usr/bin/make install' in '/home/username/ros2_ws/build/ros1_bridge'
Install the project...
-- Install configuration: ""
-- Execute custom install script
-- Symlinking: /home/username/ros2_ws/install/share/ament_index/resource_index/package_run_dependencies/ros1_bridge
-- Symlinking: /home/username/ros2_ws/install/share/ament_index/resource_index/parent_prefix_path/ros1_bridge
-- Symlinking: /home/username/ros2_ws/install/share/ros1_bridge/environment/ament_prefix_path.sh
-- Symlinking: /home/username/ros2_ws/install/share/ros1_bridge/environment/path.sh
-- Symlinking: /home/username/ros2_ws/install/share/ros1_bridge/local_setup.bash
-- Symlinking: /home/username/ros2_ws/install/share/ros1_bridge/local_setup.sh
-- Symlinking: /home/username/ros2_ws/install/share/ros1_bridge/local_setup.zsh
[...]

PR #83, opened 2017-07-24
Readme fixups

mikaelarguedas:
The first commit changes the ament invocation to use j2 to build the bridge:
 - Given that the bridge builds fine even on the Pine64 with j2. We may even we able to remove the limitation completely and provide an indication of how much RAM is used per compilation unit  and let use chose the flag matching their setup

The second commit adds `--force-cmake-configure` to the invocation given that it happens regularly that people build ROS 2, try it out and then follow the bridge instructions. At that point the package has already been built and cmake is no reran. see https://github.com/ros2/ros1_bridge/issues/82 for example

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Alternatively you can do it manually by sourcing the relevant workspaces yoursel
 Then build just the ROS 1 bridge with `-j2`:
 
 ```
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
merging this then.
Tracking the second bullet in https://github.com/ros2/ros1_bridge/issues/84

Issue #64, opened 2017-04-19
too much console output


BrannonKing:
I don't think the dynamic_bridge needs to continually spew "Passing message from ROS2 to ROS1". The message would be more useful if it included the message topic and type. However, I think we would do better to aggregate those messages and output something every five seconds saying how many of each message topic/type went which direction.

wjwwood:
Yup, the console output of the bridge could be cleaned up a lot. I think we were waiting on the console logging that @dirk-thomas is currently waiting on. That way we can make use of throttled logging messages and stuff like that. So you get some feedback without an overwhelming amount.

I'd certainly take a pr in the meantime that either disables the output or even better makes an option to enable/disable it.

mikaelarguedas:
this can now be addressed given that logging macros are implemented in [rcutils](https://github.com/ros2/rcutils/blob/master/resource/logging_macros.h.em)

allenh1:
I think a `--verbose=n` option is appropriate.

mikaelarguedas:
What does `n` refer to here ? a period in seconds ?

I think a throttled debug print would be pretty good in this scenario

allenh1:
> What does n refer to here ? 

I was referring to some sort of debug level (sorry -- that was rather unclear), but I think that a throttled print would likely be better.

But, more general still, I don't know how necessary it is to say "publishing message from ROS _ to ROS _" every single time it publishes a message.

dirk-thomas:
This should be addressed by #90 and #91.

PR #90, opened 2017-09-11
only show a message for passing the message of a specific type pair once

dirk-thomas:
The patch at least reduces the amount of printed lines (related to #64).

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -22,7 +22,11 @@
 // include ROS 1 message event
 #include "ros/message.h"
 
+#include "rcutils/logging_macros.h"
+
 #include "ros1_bridge/factory_interface.hpp"
+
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@ros2/team Waiting for review.

dirk-thomas:
I will merge this without a review in order to be used in the nightly jobs. Please feel free to comment after this has been closed.

include/ros1_bridge/factory.hpp
```diff
@@ -22,8 +22,12 @@
 // include ROS 1 message event
 #include "ros/message.h"
 
+#include "rcutils/logging_macros.h"
+
 #include "ros1_bridge/factory_interface.hpp"
 
+#define _ros1_bridge_type_to_name(x) #x
```
dirk-thomas: This never worked, see #95.

PR #91, opened 2017-09-12
remove check for header fields request_type and response_type

### LINKED ISSUE #64: too much console output

dirk-thomas:
This was introduced with the original implementation in #36.

A non C++ client might not even set these header fields. Since the bridge doesn't even use them anywhere this patch removes them (which gets rid of the repeated warning).

This PR together with #90 should resolve #64.

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -428,7 +428,7 @@ void get_ros1_service_info(
     fprintf(stderr, "%s\n", error.data());
     return;
   }
-  for (std::string field : {"type", "request_type", "response_type"}) {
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
:+1: would you mind running a packaging job with these changes to confirm everything goes as expected ?

dirk-thomas:
I ran it locally and it resolved the error message while continuing to work.

PR #96, opened 2017-10-06
print bridged type names

### LINKED ISSUE #95: Log message does not contain specific information
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - source
- Version or commit hash:
  - master
- DDS implementation:
  - doesn't matter
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Run the dynamic bridge with at least one topic pair that matches on each side, and then observe the incorrect log message.

#### Expected behavior

The log to contain something other than `ROS1_T` or `ROS2_T`.

#### Additional information

Here's an example of the log message:

```
[INFO] [ros1_bridge]: Passing message from ROS 1 ROS1_T to ROS 2 ROS2_T (showing msg only once per type) (ros1_callback() at /home/william/ros2_ws/src/ros2/ros1_bridge/include/ros1_bridge/factory.hpp:130)
```

It's nice that we went to so much effort to limit its output rate, but the content doesn't really tell you anything (neither the type nor the topic).

I believe the use of a macro here does nothing and you'd have to use `typeid(ROS1_T).name()` instead, but that will probably be mangled too:

https://github.com/ros2/ros1_bridge/blob/3f73d9a5b27f78b2a3649d938d02b1dd945bfc40/include/ros1_bridge/factory.hpp#L129-L130

dirk-thomas:
Fixes #95.

[Build Status](http://ci.ros2.org/job/ci_packaging_linux/49/)

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -26,8 +26,6 @@
 
 #include "ros1_bridge/factory_interface.hpp"
 
-#define _ros1_bridge_type_to_name(x) #x
-
 namespace ros1_bridge
 {
 
@@ -35,6 +33,12 @@ template<typename ROS1_T, typename ROS2_T>
 class Factory : public FactoryInterface
 {
 public:
+  Factory(
+    const std::string & ros1_type_name, const std::string & ros2_type_name)
+  : ros1_type_name_(ros1_type_name),
+    ros2_type_name_(ros2_type_name)
+  {}
+
   ros::Publisher
   create_ros1_publisher(
     ros::NodeHandle node,
@@ -70,7 +74,9 @@ public:
     ops.datatype = ros::message_traits::datatype<ROS1_T>();
     ops.helper = ros::SubscriptionCallbackHelperPtr(
       new ros::SubscriptionCallbackHelperT<const ros::MessageEvent<ROS1_T const> &>(
-        boost::bind(&Factory<ROS1_T, ROS2_T>::ros1_callback, _1, ros2_pub)));
+        boost::bind(
+          &Factory<ROS1_T, ROS2_T>::ros1_callback,
+          _1, ros2_pub, ros1_type_name_, ros2_type_name_)));
     return node.subscribe(ops);
   }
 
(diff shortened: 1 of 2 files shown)


include/ros1_bridge/factory.hpp
```diff
@@ -26,15 +26,21 @@
...
 template<typename ROS1_T, typename ROS2_T>
 class Factory : public FactoryInterface
 {
 public:
+  Factory(
+    const std::string & ros1_type_name, const std::string & ros2_type_name)
+  : ros1_type_name_(ros1_type_name),
+    ros2_type_name_(ros2_type_name)
+  {
+    fprintf(stderr, "Factory(%s, %s)\n", ros1_type_name_.c_str(), ros2_type_name_.c_str());
```
wjwwood: Should this be left in or removed?

include/ros1_bridge/factory.hpp
```diff
@@ -125,23 +138,25 @@ class Factory : public FactoryInterface
     auto ros2_msg = std::make_shared<ROS2_T>();
     convert_1_to_2(*ros1_msg, *ros2_msg);
     RCUTILS_LOG_INFO_ONCE_NAMED(
-      "ros1_bridge", "Passing message from ROS 1 "
-      _ros1_bridge_type_to_name(ROS1_T) " to ROS 2 "
-      _ros1_bridge_type_to_name(ROS2_T) " (showing msg only once per type)");
+      "ros1_bridge",
+      "Passing message from ROS 1 %s to ROS 2 %s (showing msg only once per type)",
+      ros1_type_name.c_str(), ros2_type_name.c_str());
```
wjwwood: Would be nice if this included the topic it was doing this on.

include/ros1_bridge/factory.hpp
```diff
@@ -125,23 +138,25 @@ class Factory : public FactoryInterface
...
   {
     ROS1_T ros1_msg;
     convert_2_to_1(*ros2_msg, ros1_msg);
     RCUTILS_LOG_INFO_ONCE_NAMED(
-      "ros1_bridge", "Passing message from ROS 2 "
-      _ros1_bridge_type_to_name(ROS2_T) " to ROS 1 "
-      _ros1_bridge_type_to_name(ROS1_T) " (showing msg only once per type)");
+      "ros1_bridge",
+      "Passing message from ROS 2 %s to ROS 1 %s (showing msg only once per type)",
+      ros1_type_name.c_str(), ros2_type_name.c_str());
```
wjwwood: Same here.

include/ros1_bridge/factory.hpp
dirk-thomas: Good catch. I removed the debug output.

include/ros1_bridge/factory.hpp
dirk-thomas: I was only aiming to fix the bug with this. Please feel free to extend the patch to add additional information.

dirk-thomas: Should we wait for this to be updated or merge it as-is and make additional improvements in a separate PR?

wjwwood: You can merge it as is, I’ll open another pr if I have a minute to add the topic name.

wjwwood: You can merge it as is, I’ll open another pr if I have a minute to add the topic name.

PR #98, opened 2017-11-17
Fix building of ros1_bridge against newer roscpp.

clalancette:
The comment in the code mostly explains it, but because of changes
to upstream catkin/roscpp, we need to expand the types of things
we look at for linking.  With this patch, building against
roscpp kinetic 1.12.7 and 1.12.12 works.

fixes ros2/build_cop#68

Signed-off-by: Chris Lalancette <clalancette@osrfoundation.org>


clalancette:
CI:

* Test packaging job [Build Status](http://ci.ros2.org/job/test_packaging_linux/10/)

PR #99, opened 2017-11-29
depend on rosidl_interfaces_packages group

mikaelarguedas:
connects to ros2/common_interfaces#45

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -32,7 +32,7 @@
   <test_depend>launch</test_depend>
   <test_depend>ros2run</test_depend>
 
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -32,6 +32,8 @@
   <test_depend>launch</test_depend>
   <test_depend>ros2run</test_depend>
 
+  <group_depend>rosidlinterfaces_packages</group_depend>
```
clalancette: I think this needs to be `rosidl_interfaces_packages`

mikaelarguedas: hmmm Looks like I was unable to type :confounded:

Issue #104, opened 2018-02-14
Bridge appears incapable of bridging ROS1 /clock to ROS2


matt-attack:
In ROS1 the rosgraph_msgs/Clock message consists of a time type internally, but in ROS2 the /clock topic is a builtin_interfaces/Time. It does not appear that the bridge can handle bridging this topic from ROS1 to ROS2.

Is there a way to force the bridge to handle this kind of conversion or should a special case be added to the bridge to allow for sim time and bag file playback?

gavanderhoorn:
Could this perhaps be solved by specifying a [custom mapping rule](https://github.com/ros2/ros1_bridge/blob/master/doc/index.rst#how-can-i-specify-custom-mapping-rule-for-messages)?

matt-attack:
Where could I put the remapping? It appears it would need to be added to the builtin_interfaces package for the bridge to accept it otherwise I get an error similar to this one:
`Ignoring rule which affects a different ROS 2 package (builtin_interfaces) then the one it is defined in (my_msgs)`

dirk-thomas:
The problem is not related to the differences in the `duration` or `time` types. Those are already being handled in the `ros`_bridge`: see https://github.com/ros2/ros1_bridge/blob/03f95fea565c2eb28057df8f7b1595bdeac995bc/include/ros1_bridge/convert_builtin_interfaces.hpp

The bridge simply can't bridge a message definition which is not available on both side. And `Clock` doesn't exist in ROS 2 atm. Please see ros2/common_interfaces#50 which add the `Clock` message for ROS 2. With that being present bridging the message will work.

dirk-thomas:
This is not a bug on the `ros1_bridge` anymore. In ROS 1 the `clock` topic is documented to have the type `rosgraph_msgs/Clock` (see http://wiki.ros.org/Clock#Clock_Message).

@tfoote Can you please follow up on this since you added the clock topic to `rclcpp`.

dirk-thomas:
I created ros2/rclcpp#473 to track this.

dirk-thomas:
Yes, you can temporarily work around it by defining a [remapping rule](https://github.com/ros2/ros1_bridge/blob/master/doc/index.rst#how-can-i-specify-custom-mapping-rule-for-messages).

PR #106, opened 2018-03-26
Map duration and time messages

dirk-thomas:
Only slightly related to ros2/ros1_bridge#104.

While the bridge can already handle fields with a `Duration` or `Time` type it currently can't map the messages `std_msgs/Duration|Time` with `builtin_interfaces/Duration|Time`.

With this patch the following appears when invoking the dynamic bridge with `--print-pairs`:

```
  - 'builtin_interfaces/Duration' (ROS 2) <=> 'std_msgs/Duration' (ROS 1)
  - 'builtin_interfaces/Time' (ROS 2) <=> 'std_msgs/Time' (ROS 1)
```

And the following example works:

* ROS 1 side:
  ```
  rostopic pub /time std_msgs/Time "data:
    secs: 1 
    nsecs: 2"
  ```
* ROS 2 side:
  ```ros2 topic echo /time builtin_interfaces/Time```

### DIFF
--- a/include/ros1_bridge/convert_builtin_interfaces.hpp
+++ b/include/ros1_bridge/convert_builtin_interfaces.hpp
@@ -31,27 +31,26 @@ namespace ros1_bridge
 template<>
 void
 convert_1_to_2(
-  const ros::Duration & ros1_msg,
+  const ros::Duration & ros1_type,
   builtin_interfaces::msg::Duration & ros2_msg);
 
 template<>
 void
 convert_2_to_1(
   const builtin_interfaces::msg::Duration & ros2_msg,
-  ros::Duration & ros1_msg);
-
+  ros::Duration & ros1_type);
 
 template<>
 void
 convert_1_to_2(
-  const ros::Time & ros1_msg,
+  const ros::Time & ros1_type,
   builtin_interfaces::msg::Time & ros2_msg);
 
 template<>
 void
 convert_2_to_1(
   const builtin_interfaces::msg::Time & ros2_msg,
-  ros::Time & ros1_msg);
+  ros::Time & ros1_type);
 
 }  // namespace ros1_bridge
 
--- a/resource/get_factory.cpp.em
+++ b/resource/get_factory.cpp.em
@@ -19,6 +19,7 @@
 from ros1_bridge import camel_case_to_lower_case_underscore
 }@
 #include "ros1_bridge/factory.hpp"
+#include "ros1_bridge/builtin_interfaces_factories.hpp"
 
 @[for ros2_package_name in sorted(ros2_package_names)]@
 #include "@(ros2_package_name)_factories.hpp"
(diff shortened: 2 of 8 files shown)


dirk-thomas:
Waiting for review.

include/ros1_bridge/convert_builtin_interfaces.hpp
```diff
@@ -31,27 +31,26 @@ namespace ros1_bridge
 template<>
 void
 convert_1_to_2(
-  const ros::Duration & ros1_msg,
+  const ros::Duration & ros1_type,
```
dhood: This change doesn't make sense to me. Can you clarify why you changed only one of the names of this prototype? and that the names don't match what's used in the implementation at https://github.com/ros2/ros1_bridge/pull/106/files#diff-e7604d29ef87753b67dcf843c9e940caR72?

mikaelarguedas: these are not messages but ROS 1 types hence the change in name. The prototype you're pointing to is taking messages and not types so using the `_msg` there is appropriate

dirk-thomas: This patch changes the `ros1_msg` argument to `ros1_type` argument for all `convert_1_to_2` / `convert_2_to_1` functions in the header as well as the source file. The rational is that neither of the two arguments `ros::Duration` nor `ros::Time` is a message but just a plain type.

The function you are referencing is unrelated to this signature. That one is for actual message in the `builtin_interfaces` packages and therefore those arguments still use `_msg` as the argument suffix.

resource/get_factory.cpp.em
```diff
@@ -18,8 +18,8 @@
 @{
 from ros1_bridge import camel_case_to_lower_case_underscore
 }@
-#include "ros1_bridge/factory.hpp"
 #include "ros1_bridge/builtin_interfaces_factories.hpp"
+#include "ros1_bridge/factory.hpp"
```
dirk-thomas: I would rather revert this change. Before it was including first the factory interface and then all specialized factories. That made sense to me. Also the include order isn't being checked at the moment since some includes are hard coded and some other are being generated. So the overall order won't pass the inter anyway (without custom logic ordering the superset).

mikaelarguedas: Feel free to revert it, I wanted to save an iteration but apparently I should have just commented to get feedback rather than commiting the change

dirk-thomas: Np, better being proactive - I appreciate it!

include/ros1_bridge/convert_builtin_interfaces.hpp
dhood: ok, I'll keep looking at this to get my head around it myself, in the meantime the approval from @mikaelarguedas suggests it's just my misunderstanding so not a blocker :smile:

dhood: following up, @mikaelarguedas pointed out what I was missing, which is that these particular msgs, fundamentally, are represented differently in ROS 1 and ROS 2: time/duration isn't a message in ROS 1, it is a data type, whereas in ROS 2 time/duration has message definitions associated with them. The contents of `ros1_msg.data` at [this line](https://github.com/ros2/ros1_bridge/pull/106/files#diff-e7604d29ef87753b67dcf843c9e940caR75) is not another message.

PR #110, opened 2018-04-03
Fix linter failures from includes

dhood:
followup of https://github.com/ros2/ros1_bridge/pull/106

example failures: [Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux/1027/)

CI: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/71/)

The ros1 headers have to come first to avoid "Found C system header after other header. Should be: builtin_interfaces_factories.h, c system, c++ system, other."

Then I tried to keep ros1 and ros2 message includes together, then the others. I know there's a strange mix of system headers, it's what was needed to make the linters pass. If you prefer a different combination please commit directly after making sure it passes the linters yourself.

### DIFF
--- a/include/ros1_bridge/builtin_interfaces_factories.hpp
+++ b/include/ros1_bridge/builtin_interfaces_factories.hpp
@@ -15,8 +15,6 @@
 #ifndef ROS1_BRIDGE__BUILTIN_INTERFACES_FACTORIES_HPP_
 #define ROS1_BRIDGE__BUILTIN_INTERFACES_FACTORIES_HPP_
 
-#include <ros1_bridge/factory.hpp>
-
 // include ROS 1 messages
 #include <std_msgs/Duration.h>
 #include <std_msgs/Time.h>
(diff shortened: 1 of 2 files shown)


include/ros1_bridge/builtin_interfaces_factories.hpp
```diff
@@ -25,11 +23,18 @@
 #include <builtin_interfaces/msg/duration.hpp>
 #include <builtin_interfaces/msg/time.hpp>
 
+#include <memory>
+#include <string>
```
dirk-thomas: Can these two be at the top of the file?

src/builtin_interfaces_factories.cpp
```diff
@@ -12,17 +12,22 @@
...
 
+#include <memory>
+#include <string>
+
 #include "rclcpp/rclcpp.hpp"
 
 // include builtin interfaces
-#include <ros1_bridge/builtin_interfaces_factories.hpp>
-#include <ros1_bridge/convert_builtin_interfaces.hpp>
+#include "ros1_bridge/convert_builtin_interfaces.hpp"
```
mikaelarguedas: did `cpplint` ask to swap these 2 ?

mikaelarguedas:
yeah cpplint is pretty dumb when it comes to include order. Thanks @dhood for catching and following up on this!

include/ros1_bridge/builtin_interfaces_factories.hpp
dhood: they cannot (see description)

dhood:
I tried many combinations that seemed more reasonable and this is what stuck. If you find a nice combo that passes the linters feel free to commit it

include/ros1_bridge/builtin_interfaces_factories.hpp
dirk-thomas: Sorry, didn't read to closely. :+1:

PR #112, opened 2018-04-08
Duplicate messages in bidirectional_bridge fix

ArkadiuszNiemiec:
Issue mentioned in #43 and #111, solution based on discussion in #43. Branched from ```ardent``` as I am not able to build ```master``` on my system.

Tested by using ```parameter_bridge``` and multiple ROS1 and ROS2 publishers and subscribers on the same topic.

Thank you @wjwwood for your help!

Issue #111, opened 2018-04-05
Bidirectional bridge duplicates messages in ROS1


Ekci:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - from source
- Version or commit hash:
  - 0.4.0
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
I have generated a bidirectional bridge using parameter_bridge to send a std_msgs/Float32MultiArray messages from ROS1 to ROS2. It works but I have just realized that when I use ```rostopic hz /topic``` I get twice the frequency I set in my code. Turning off bridge make ```rostopic hz /topic``` show correct value. Using ```rostopic info /topic``` shows:

```Publishers: 
 * /start_controller (http://xxx:44737/)
 * /ros_bridge (http://xxx:38531/)

Subscribers: 
 * /rostopic_28342_1522934521725 (http://xxx:46713/)
 * /ros_bridge (http://xxx:38531/)
```

Inspecting ROS1 topic clearly shows that every message is duplicated. This makes me think that parameter_bridge sends ROS1 message to ROS2 and then from ROS2 back to ROS1. Unfortunately it's not looping back and doesn't crash: messages on ROS2 side are not duplicated.

dirk-thomas:
Duplicate of #43.

dhood:
@Ekci thanks for reporting this. As mentioned, there's useful info in #43, including pointers if you are interested in working on a fix for the issue yourself (in which case we would welcome a pull request)

Issue #43, opened 2016-12-06
Bidirectional bridge duplicating messages with FastRTPS


vilhjalmur89:
I've been trying out ROS 2 and I'm seeing unexpected behavior when using the bridge. 

The following image shows three messages being sent; First without any bridge, then with a FastRTPS-bridge and finally with a OpenSplice-bridge. 

The static_bridge with FastRTPS passes a message from Ros 1 to Ros 2, but also back from ROS 2 to ROS 1, even though there was no ROS 2 process running (except for the bridge).
The same scenario with OpenSplice only passes the message from ROS 1 to ROS 2, as expected. 

![static_bridge_fastrtps_and_opensplice](https://cloud.githubusercontent.com/assets/5200924/20923063/f1e6b154-bbaa-11e6-96ab-72e5e9542331.png)

dhood:
Thank you for the issue report @vilhjalmur89 

I can confirm this issue happens for both `static_bridge` and `simple_bridge`. With the `simple_bridge` you get some additional logging that shows this in action:

```
dhood@osrf-esteve:~ [~]$ simple_bridge__rmw_opensplice_cpp 
I heard from ROS 1 from: [/rostopic_17524_1481068220810]
I heard from ROS 1: [static_bridge__rmw_opensplice_cpp is running]
Passing along to ROS 2: [static_bridge__rmw_opensplice_cpp is running]
^Csignal_handler(2)
dhood@osrf-esteve:~ [~]$ simple_bridge__rmw_fastrtps_cpp
I heard from ROS 1 from: [/rostopic_17588_1481068239197]
I heard from ROS 1: [static_bridge__rmw_fastrtps_cpp is running]
Passing along to ROS 2: [static_bridge__rmw_fastrtps_cpp is running]
  I heard from ROS 2: [static_bridge__rmw_fastrtps_cpp is running]
  Passing along to ROS 1: [static_bridge__rmw_fastrtps_cpp is running]
    I heard from ROS 1 from myself
```

[...]

PR #113, opened 2018-04-11
Duplicate messages in bidirectional_bridge fix

ArkadiuszNiemiec:
Issue mentioned in #43 and #111, solution based on discussion in #43. Branched from ardent as I am not able to build master on my system.

Tested by using parameter_bridge and multiple ROS1 and ROS2 publishers and subscribers on the same topic.

Thank you @wjwwood for your help!

This PR is copy of broken PR #112. Sorry for the inconvenience.

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -85,7 +85,8 @@ public:
     rclcpp::Node::SharedPtr node,
     const std::string & topic_name,
     size_t queue_size,
-    ros::Publisher ros1_pub)
+    ros::Publisher ros1_pub,
+    rclcpp::PublisherBase::SharedPtr ros2_pub = nullptr)
   {
     rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
     custom_qos_profile.depth = queue_size;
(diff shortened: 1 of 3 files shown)


ArkadiuszNiemiec:
@wjwwood It now passes all tests.

include/ros1_bridge/factory.hpp
```diff
@@ -145,10 +149,22 @@ class Factory : public FactoryInterface
...
+    rclcpp::PublisherBase::SharedPtr ros2_pub = nullptr)
   {
+    if (ros2_pub) {
+      bool result = false;
+      auto ret = rmw_compare_gids_equal(&msg_info.publisher_gid, &ros2_pub->get_gid(), &result);
+      if (ret == RMW_RET_OK) {
+        if (result) {  // message GID equals to bridge's ROS2 publisher GID
+          return;  // do not publish messages from bridge itself
+        }
+      }
```
wjwwood: The else case should be handled here. If `rmw_compare_gids_equal` fails, then we should probably raise an exception or at least log an error message to the console using `RCLCPP_ERROR` logging macro.

wjwwood: You can get the underlying error message with `rmw_get_error_string_safe()`, and after getting it you should clear the error state with `rmw_reset_error()`.

ArkadiuszNiemiec: Ok, I have found similar code on different repo. Added.

ArkadiuszNiemiec:
> I still haven't reproduced it myself, but I assume you have seen that it fixes your original issue?

Yes, I have tested it by starting two nodes publishing on one topic. After including the fix the duplicates went away. I have started ROS2 node on the same topic and it also bridged perfectly.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4202/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1259/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3484/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4319/)

wjwwood:
@ArkadiuszNiemiec looks like the linters are still unhappy about some trailing whitespace or something. Can you check `ament_uncrustify` and `ament_cpplint` one more time?

ArkadiuszNiemiec:
@wjwwood Fixed, empty line was not empty :smile:

wjwwood:
Thanks @ArkadiuszNiemiec!

PR #124, opened 2018-06-18
print service pairs as well

### LINKED ISSUE #115: --print-pairs prints only messages and not services
I think it's a bug but maybe this is on purpose ?

dirk-thomas:
Fixes #115. Connect to #115.


dirk-thomas:
> be good to update the template header to list the new variable used

Good point - added in 9bda2b2f1231bc088652b22c2d49c6f5e1b72252.

PR #138, opened 2018-10-04
Changed cmake message logger level

alsora:
When building this package using 

`colcon build`

No message is displayed if the ros1 libraries are not found, due to the `STATUS` logger level of these CMake messages. 
Changing the logger level to `WARNING` solves the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ include(cmake/find_ros1_package.cmake)
 
 find_package(PkgConfig)
 if(NOT PKG_CONFIG_FOUND)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Looks like a good idea. Thanks for the patch.

clalancette:
The problem with this change is that it now triggers warning detection in CI.  See, for example, https://ci.ros2.org/job/ci_linux/5266/warnings2Result/new/ , where the build is marked UNSTABLE because of the output here.

dirk-thomas:
Good point, blacklisted for all jobs which don't have ROS 1 available anyway: ros2/ci/pull/215

PR #140, opened 2018-10-23
Add a whitelisting dynamic bridge 

dhananjaysathe:
This is a fusion of ideas from the parameter bridge and dynamic bridge.
It add the ability to set a list of regular expressions for topics and services to be bridged.
This is often useful when you want the fancy feature of dynamic bridge automatically bridging ROS1<=>ROS2 services and topics but want to limit it to a few topics (often machine generated hence the need for a regex)


include/ros1_bridge/factory.hpp
```diff
@@ -128,9 +130,10 @@ class Factory : public FactoryInterface
       return;
     }
 
-    std::string key = "callerid";
-    if (connection_header->find(key) != connection_header->end()) {
-      if (connection_header->at(key) == "/ros_bridge") {
+    const auto caller = connection_header->find("callerid");
```
dhananjaysathe: This stuff is need to allow for a random name for the bridge that makes it easier to identify

PR #143, opened 2018-11-13
Fix for actions subfolder introduction in ros2 message bridge

sservulo:
Relates to the change https://github.com/ros2/rosidl/commit/7d791728f36b99f08242aedd33aac05f95c64619 pointed out by @mjcarroll and @dirk-thomas which introduced packaging failures in `ros1_bridge`.

Deprecates https://github.com/ros2/rosidl/pull/319 .

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -203,7 +203,11 @@ def get_ros2_messages():
         pkgs.append(package_name)
         resource, _ = ament_index_python.get_resource(resource_type, package_name)
         interfaces = resource.splitlines()
-        message_names = [i[:-4] for i in interfaces if i.endswith('.msg')]
+        message_names = [n[4:-4] for n in interfaces if n.startswith('msg/') and n.endswith('.msg')]
+        message_names.extend(
+            [n[4:-4] for n in interfaces if n.startswith('srv/') and n.endswith('.msg')])
+        message_names.extend(
+            [n[7:-4] for n in interfaces if n.startswith('action/') and n.endswith('.msg')])
         for message_name in message_names:
             msgs.append(Message(package_name, message_name, prefix_path))
         # check package manifest for mapping rules


mjcarroll:
Linux packaging: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/144/)

mjcarroll:
Looks like the tests are still timing out on the farm.

dirk-thomas:
@sservulo Please build the branch locally and run the tests.

mjcarroll:
Yes, there are some `flake8` errors, and it looks like service pairs aren't being correctly generated.

ros1_bridge/__init__.py
```diff
@@ -659,9 +669,13 @@ def load_ros1_service(ros1_srv, rospack=None):
...
-    message_path = os.path.join(
-        ros2_msg.prefix_path, 'share', ros2_msg.package_name, 'msg',
-        ros2_msg.message_name + '.msg')
+    message_path = None
+    for message_location in ['msg', 'srv', 'action']:
+        message_path = os.path.join(
+            ros2_msg.prefix_path, 'share', ros2_msg.package_name, message_location,
+            ros2_msg.message_name + '.msg')
+        if os.path.isfile(message_path):
+            break
```
dirk-thomas: To different type of interfaces with the same name can exist. So only allowing one here seems wrong.

Same below.

hidmic: Just to double check, this PR is specifically about stripping subfolders from interface files, not to provide any support for bridging actions, right? If that's the case, then it seems to me we don't need to expose messages generated to implement services nor messages and services generated to implement actions.

apojomovsky:
Linux packaging: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/147/)

mjcarroll:
For good measure:

* Windows packaging: [Build Status](https://ci.ros2.org/job/ci_packaging_windows/82/)
* MacOS packaging: [Build Status](https://ci.ros2.org/job/ci_packaging_osx/57/)

sloretz:
Windows packaging job has warnings.

I don't think the warning script is catching this one, not sure how long it's been present.
```
warning: install_lib: 'C:\J\workspace\ci_packaging_windows\ws\build\test_launch_ros\build\lib' does not exist -- no Python modules to install
```

and 
```
c:\j\workspace\ci_packaging_windows\ws\install\include\urdf_model\color.h(79): warning C4244: 'argument': conversion from 'double' to 'const float', possible loss of data [C:\J\workspace\ci_packaging_windows\ws\build\robot_state_publisher\robot_state_publisher_solver.vcxproj
```

which might be caused by ros/urdfdom_headers#47/

mjcarroll:
I don't think either of those are a blocker, since they are from unrelated packages.

PR #146, opened 2018-11-21
expose convert function

Karsten1987:
In order to leverage the bridge to convert ros messages from ros1 to ros2, we would like to expose this function. This allows us to convert messages without using pub/sub.

cc @botteroa-si

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,8 +63,6 @@ endif()
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 
-ament_export_dependencies(roscpp)
-
 ament_python_install_package(${PROJECT_NAME})
 
 ament_package()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,8 @@ endif()
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 
+ament_export_dependencies(roscpp)
+
 ament_python_install_package(${PROJECT_NAME})
 
 ament_package()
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -63,6 +63,8 @@ endif()
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 
+ament_export_dependencies(roscpp)
```
dirk-thomas: This isn't going to work the way intended. `roscpp` is a ROS 1 package and should be found using the custom CMake function used above (`find_ros1_package` which doesn't rely on CMake config files but `pkg-config` instead). If you want this to be exposed in downstream package it would need to be done in an extra file by invoking `find_ros1_package`.

botteroa-si: Hi @dirk-thomas, our problem is the following: we link against the ros1_bridge and on Linux everything works fine, whereas on Mac the linker cannot find `roscpp` symbols.
We could solve this by finding `roscpp` ourselves, as you said, but for us it is only a transitive dependency for the bridge, therefore we thought that this would be the best solution: with this change, the `roscpp` dependency is exported and everything works as expected.

dirk-thomas: I understand your use case but this is still the wrong approach. As soon as a `roscpp` package is created in ROS 2 e.g. to provide an API shim the current approach is going to fail. The `ros1_bridge` package needs to invoke `find_ros1_package` in an extra file instead.

botteroa-si: Ok, we will then use the `find_ros1_package` macro in our package.

dirk-thomas: I am not saying you have call it in your package. The logic can be placed in a CMake extra file in `ros2_bridge` in order to avoid having to repeat it in multiple downstream packages.

botteroa-si:
@dirk-thomas, that would of course be the cleanest solution, but currently we don't have time to look into it. Therefore, for the moment, we would just call the `find_ros1_package()` in our package. 
Thank you for your comments! :)

PR #147, opened 2018-11-21
support for custom field mapping for services

dirk-thomas:
Similar functionality as it is already possible for messages.


dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/148/)

PR #152, opened 2018-12-10
exclude ros1 nodelets

Karsten1987:
connects to ros2/rosbag2#69
this PR excludes nodelets so that rosbag2 can replay ros1 bag files without linking conflicts.


Karsten1987:
CI packaging job to test the bridge
[Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux/1283/)

dirk-thomas:
:+1: This will fix the imminent problem and not link against `libnodelet`.

I am wondering though if this works if ROS 1 is built in isolation. Are the services in the `nodelet` package just not available for bridging or does it fail anywhere due to no finding their headers?

CMakeLists.txt
```diff
@@ -45,7 +45,9 @@ find_ros1_interface_packages(ros1_message_packages)
 set(prefixed_ros1_message_packages "")
 foreach(ros1_message_package ${ros1_message_packages})
   find_ros1_package(${ros1_message_package} REQUIRED)
```
Martin-Idel-SI: I think this should also be done within the `if(NOT ...)` call - we don't need to find the package if we don't use it.

Karsten1987: done in https://github.com/ros2/ros1_bridge/pull/152/commits/6c953ce386ac8c6785e3c03472ba9d0b6a98bfe8

Karsten1987:
[Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux/1285/)

wjwwood:
The _right_ solution to this problem is to ensure that the class loader used in ROS 1 and ROS 2 can be used at the same time, either by using different namespaces or by hiding conflicting symbols so they don't collide at link time or by making it so that the same exact version of class loader is used in both.

Karsten1987:
I went ahead and merged this. I've put a TODO in the CMakeLists and opened an issue on ros/class_loader.
I wasn't entirely sure whether class_loader is the right repo over ros/pluginlib.

PR #153, opened 2018-12-11
add support for services in parameter_bridge and fixed-size array in service definitions

cyrilleberger:
This pull request has two commits:

1. add support for fixed size array in service calls (I have tried to imitate what was doing for topics)
2. making possible to define services using the parameter_bridge, it is done very similarly to topics:
  It needs to parameter 'services_1_to_2' (for calling ROS2 services from ROS1) and 'services_2_to_1 (for calling ROS1 from ROS2), I tried to follow the naming convention of the dynamic_bridge. The parameters should look like:
    [
      {service: /name/of/service package: name_of_package_with_service, type: type_of_message},
      ...
    ]
For calling ROS1 add_two_ints from ROS2:
    [
      {service: /add_two_ints package: roscpp_tutorials, type: TwoInts}
    ]

Sorry for having two features in one pull request, it is hard for me to test one without the other, if you prefer separate pull request before inclusion, I will do that when patches are ready. Since master does not build with bouncy, I have targeted bouncy branch, I suspect it works with master as well, but when Crystal is released, I will test with that branch.


dirk-thomas:
Thank you for working on this. Please retarget `master` since new features like this won't be released into Bouncy.

cyrilleberger:
Updated in #158 for master and crystal.

cyrilleberger:
Example of srvs file with fixed size array: https://github.com/cyrilleberger/fixed_size_array_msgs

PR #158, opened 2019-01-14
add support for services in parameter_bridge and fixed-size array in service definitions

cyrilleberger:
This is a new version of #153 (I couldn't update the old pull request to master)

This pull request has two commits:

1. add support for fixed size array in service calls (I have tried to imitate what was doing for topics)
2. making possible to define services using the parameter_bridge, it is done very similarly to topics:
  It needs to parameter 'services_1_to_2' (for calling ROS2 services from ROS1) and 'services_2_to_1 (for calling ROS1 from ROS2), I tried to follow the naming convention of the dynamic_bridge. The parameters should look like:
    [
      {service: /name/of/service package: name_of_package_with_service, type: type_of_message},
      ...
    ]
For calling ROS1 add_two_ints from ROS2:
    [
      {service: /add_two_ints package: roscpp_tutorials, type: TwoInts}
    ]

Sorry for having two features in one pull request, it is hard for me to test one without the other, if you prefer separate pull request before inclusion, I will do that when patches are ready.

I have tested compilation with Crystal, run tests are under way.

### DIFF
--- a/resource/pkg_factories.cpp.em
+++ b/resource/pkg_factories.cpp.em
@@ -235,7 +235,17 @@ void ServiceFactory<
 ) {
 @[      for field in service["fields"][type.lower()]]@
 @[        if field["array"]]@
+@[          if field["dynamic_array"]]@
+  // ensure array size
+@[            if field["upper_bound_array"] > 0]@
+  // check boundary
+@[              if frm == "1"]@
+                  assert(req@(frm).@(field["ros1"]["name"]).size() <= @(field["upper_bound_array"]));
+@[              end if]@
+@[            end if]@
+  // dynamic arrays must be resized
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> This pull request has two commits:

Since both features are unrelated to each other I would suggest to create two separate PRs for them. That allows to iterate on them independently and not block one due to pending items on the other and it will keep the discussion focused on one topic.

dirk-thomas:
@cyrilleberger Please split the PR as mentioned before in order to move forward with it.

dirk-thomas:
@cyrilleberger Friendly ping.

cyrilleberger:
Split up in #175 and #176 (sorry for the delay, I have been too busy :()

PR #160, opened 2019-01-23
Allow external use of ros1_bridge factories

paulbovbel:
Expose libros1_bridge functions via headers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ custom_executable(simple_bridge "src/simple_bridge.cpp"
 add_library(${PROJECT_NAME} SHARED
   "src/builtin_interfaces_factories.cpp"
   "src/convert_builtin_interfaces.cpp"
+  "src/bridge.cpp"
   ${generated_files})
 ament_target_dependencies(${PROJECT_NAME}
   ${prefixed_ros1_message_packages}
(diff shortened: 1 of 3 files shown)


include/ros1_bridge/bridge.hpp
```diff
@@ -69,7 +69,10 @@ get_factory(
   const std::string & ros1_type_name,
   const std::string & ros2_type_name);
 
-Bridge1to2Handles
+std::unique_ptr<ServiceFactoryInterface>
+get_service_factory(const std::string &, const std::string &, const std::string &);
+
+inline Bridge1to2Handles
```
dirk-thomas: Why should all these functions be changed to `inline`?

paulbovbel: thanks for the patience:

since these methods are implemented in the definition, an `inline` is a simple way to avoid `multiple definition of` linker errors.

dirk-thomas: I would suggest to move the definition of the functions into a `.cpp` file then.

paulbovbel: done!

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/163/)

dirk-thomas:
Please address the linter warning about the missing include.

paulbovbel:
Done

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/165/)

dirk-thomas:
Thanks for the improvement.

PR #162, opened 2019-02-06
Allow custom QoS settings for pub/sub factories

paulbovbel:
This allows specifying a custom QoS profile for ROS2 components, or setting a `latch` flag on the created ROS1 publisher.

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -45,9 +45,10 @@ public:
   create_ros1_publisher(
     ros::NodeHandle node,
     const std::string & topic_name,
-    size_t queue_size)
+    size_t queue_size,
+    bool latch = false)
   {
-    return node.advertise<ROS1_T>(topic_name, queue_size);
+    return node.advertise<ROS1_T>(topic_name, queue_size, latch);
   }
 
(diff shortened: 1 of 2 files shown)


include/ros1_bridge/factory.hpp
```diff
@@ -45,9 +45,10 @@ class Factory : public FactoryInterface
   create_ros1_publisher(
     ros::NodeHandle node,
     const std::string & topic_name,
-    size_t queue_size)
+    size_t queue_size,
+    bool latch)
```
dirk-thomas: Why does the API not match the interface - with the `latch` argument being optional?

Same below for the other optional arguments.

paulbovbel: fixed, thanks

paulbovbel:
anything else I should adjust?

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/166/)

dirk-thomas:
@paulbovbel This PR introduced API with an argument type `rmw_qos_profile_t`. Is there a specific reason for that or could code using this API be updated to use a similar API with an argument type `rclcpp::QoS`?

PR #164, opened 2019-02-13
export find_ros1_package cmake

Karsten1987:
Signed-off-by: Karsten Knese <karsten@openrobotics.org>

I believe this is the correct fix for exporting the `find_ros1_package.cmake. This allows other packages to use this macro straight after finding the `ros1_bridge` package.
see: https://github.com/ros2/rosbag2/pull/90#discussion_r256232380

please advise how to test this patch. Also I would appreciate if this could be backported to crystal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,9 @@ find_ros1_package(roscpp)
 if(NOT ros1_roscpp_FOUND)
   message(WARNING "Failed to find ROS 1 roscpp, skipping...")
   # call ament_package() to avoid ament_tools treating this as a plain CMake pkg
-  ament_package()
+  ament_package(
+    CONFIG_EXTRAS cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake
+  )
   return()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,9 @@ ament_export_libraries(${PROJECT_NAME})
 
 ament_python_install_package(${PROJECT_NAME})
 
-ament_package()
+ament_package(
+  CONFIG_EXTRAS cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake
+)
 
 set(generated_path "${CMAKE_BINARY_DIR}/generated")
 set(generated_files "${generated_path}/get_factory.cpp")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ find_package(rmw_implementation_cmake REQUIRED)
 find_package(std_msgs REQUIRED)
 
 # find ROS 1 packages
-list(APPEND ros1_cmake_macros cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake)
+set(cmake_extras_files cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake)
 include(cmake/find_ros1_package.cmake)
 
 find_package(PkgConfig)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> please advise how to test this patch.

You can test it with the referenced `rosbag2` PR which shouldn't need the `include` calls anymore.

Karsten1987:
in order to compile these packages, the `ros1_bridge` has to be compiled. Which CI job could do that?

dirk-thomas:
https://ci.ros2.org/job/ci_packaging_linux/

Karsten1987:
In order to use the cmake macros, the `pkg-config` dependency has to be exported with it.
I am not sure, but when interpreting [rep140](http://www.ros.org/reps/rep-0140.html#build-export-depend-multiple) correctly, the `pkg-config` dependency should be exported via `build_export_depend`. Please advise if not.

dirk-thomas:
`buildtool_export_depend` - "tool" because `pkg-config` needs to be present on the host system rather than on the target system in case of cross compilation.

Karsten1987:
comments are addressed and CI was setup here: [Build Status](https://ci.ros2.org/view/All/job/linux_packaging_post_rosbridge/21/)

putting this in review

cmake/find_ros1_package.cmake
```diff
@@ -20,6 +20,14 @@ macro(find_ros1_package name)
       "${ARG_UNPARSED_ARGUMENTS}")
   endif()
 
+  find_package(PkgConfig)
+  if(NOT PKG_CONFIG_FOUND)
+    message(WARNING "Failed to find PkgConfig, skipping...")
+    # call ament_package() to avoid ament_tools treating this
+    # as a plain CMake pkg
+    ament_package()
+    return()
```
dirk-thomas: This CMake extra file can't call `ament_package()`. Instead the `find_package(PkgConfig)` call should be `REQUIRED`.

Karsten1987:
new CI: [Build Status](https://ci.ros2.org/view/All/job/linux_packaging_post_rosbridge/22/)

CMakeLists.txt
```diff
@@ -69,7 +71,9 @@ ament_export_libraries(${PROJECT_NAME})
 
 ament_python_install_package(${PROJECT_NAME})
 
-ament_package()
+ament_package(
+  CONFIG_EXTRAS cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake
```
dirk-thomas: Nitpick: maybe move the duplicated arguments into a variable?

paulbovbel:
Having made use of this cmake snippet elsewhere in the past, ros1_bridge is a very heavy dependency size/build time wise) to pull in to get access to this cmake helper, while the functionality is very helpful anywhere you want to pull catkin assets into an ament package.

Would it make sense to pull apart ros1_bridge repository into several packages? IMO helpers like this could live in a `ros1_interface` or `ros1_tools`-like package.

dirk-thomas:
> Would it make sense to pull apart ros1_bridge repository into several packages (before dependencies on ros1_bridge are established, and unable to be broken)? IMO helpers like this could live in a `ros1_interface` or `ros1_tools`-like package.

That would absolutely be feasible. (I wouldn't use `interface` in the package name since that usually indicates messages / services / actions - especially with the move to IDL) Please feel free to create a package like that and propose a PR to this package to use the functionality from there. Please be aware that a package using both ROS 2 and ROS 1 might not be supported out-of-the-box on the CI infrastructure though so I would suggest to use this very sparsely.

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ find_package(rmw_implementation_cmake REQUIRED)
 find_package(std_msgs REQUIRED)
 
 # find ROS 1 packages
+list(APPEND ros1_cmake_macros cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake)
```
dirk-thomas: Since the variable `ros1_cmake_macros` doesn't exist yet use `set()` instead. Based on the usage I would also suggest to name the variable `cmake_extra_files`.

Karsten1987:
latest CI: [Build Status](https://ci.ros2.org/view/All/job/linux_packaging_post_rosbridge/27/)

merging :)

PR #175, opened 2019-03-21
add support for fixed size array in service calls

cyrilleberger:
add support for fixed size array in service calls (I have tried to imitate what was doing for topics)


ros1_bridge/__init__.py
```diff
@@ -538,24 +538,27 @@ def determine_common_services(ros1_srvs, ros2_srvs, mapping_rules):
...
                 ros1_type = ros1_field[0]
-                ros2_type = str(ros2_fields[direction][i].type)
+                ros2_type = ros2_fields[direction][i].type
                 ros1_name = ros1_field[1]
                 ros2_name = ros2_fields[direction][i].name
-                if ros1_type != ros2_type or ros1_name != ros2_name:
+                if ros1_type != str(ros2_type) or ros1_name != ros2_name:
                     match = False
                     break
+                ros2_cpptype = ros2_type.pkg_name + "::msg::" + ros2_type.type if ros2_type.pkg_name else ""
```
dirk-thomas: This might work for now but for actions the namespace of fields might not be `msg`.

Please use single quotes.

ros1_bridge/__init__.py
```diff
@@ -538,24 +538,27 @@ def determine_common_services(ros1_srvs, ros2_srvs, mapping_rules):
...
                 output[direction].append({
-                    'basic': False if '/' in ros1_type else True,
-                    'array': True if '[]' in ros1_type else False,
+                    'basic': not ros2_type.pkg_name,
+                    'array': ros2_type.is_array,
+                    'dynamic_array': not ros2_type.array_size and not ros2_type.is_upper_bound,
```
dirk-thomas: This should only be `True` when `ros2_type.is_array` is `True`.

ros1_bridge/__init__.py
```diff
@@ -538,24 +538,27 @@ def determine_common_services(ros1_srvs, ros2_srvs, mapping_rules):
...
+                    'upper_bound_array': ros2_type.array_size if ros2_type.is_upper_bound else -1,
```
dirk-thomas: The default value should be `None` instead.

resource/pkg_factories.cpp.em
```diff
@@ -235,7 +235,17 @@ void ServiceFactory<
 ) {
 @[      for field in service["fields"][type.lower()]]@
 @[        if field["array"]]@
+@[          if field["dynamic_array"]]@
+  // ensure array size
+@[            if field["upper_bound_array"] > 0]@
+  // check boundary
+@[              if frm == "1"]@
```
dirk-thomas: The two conditions from line 240 and 242 can be merged into a single `if`.

ros1_bridge/__init__.py
cyrilleberger: No idea what to do about the '::msg::', should it be an argument to the function? (that function seems to be all about services)

cyrilleberger: Actually, thinking about it, isn't this always 'msg' even for actions since it is the type of a field?

ros1_bridge/__init__.py
```diff
@@ -538,24 +538,27 @@ def determine_common_services(ros1_srvs, ros2_srvs, mapping_rules):
...
+                if ros1_type != str(ros2_type) or ros1_name != ros2_name:
                     match = False
                     break
+                ros2_cpptype = ros2_type.pkg_name + '::msg::' + ros2_type.type if ros2_type.pkg_name else ''
                 output[direction].append({
-                    'basic': False if '/' in ros1_type else True,
-                    'array': True if '[]' in ros1_type else False,
+                    'basic': not ros2_type.pkg_name,
+                    'array': ros2_type.is_array,
+                    'dynamic_array': ros2_type.is_array and not ros2_type.array_size and not ros2_type.is_upper_bound,
```
dirk-thomas: Based on the above changes to the template I assume you want `dynamic_array` to be `True` for bounded as well as unbounded arrays. Then the condition needs fixing:

`ros2_type.is_array and (not ros2_type.array_size or ros2_type.is_upper_bound)`

resource/pkg_factories.cpp.em
```diff
@@ -235,7 +235,15 @@ void ServiceFactory<
 ) {
 @[      for field in service["fields"][type.lower()]]@
 @[        if field["array"]]@
+@[          if field["dynamic_array"]]@
+  // ensure array size
+@[            if field["upper_bound_array"] != None && frm == "1"]@
```
dirk-thomas: Should be: `is not None`.

ros1_bridge/__init__.py
dirk-thomas: Even though the field is a message it might still be in the `action` namespace.

resource/pkg_factories.cpp.em
```diff
@@ -39,7 +39,7 @@ from ros1_bridge import camel_case_to_lower_case_underscore
 namespace ros1_bridge
 {
 
-std::shared_ptr<FactoryInterface>
+std::shared_ptr<FactoryInterface>git push --force-with-lease origin services-static-bridge
```
dirk-thomas: Something is wrong here...

cyrilleberger: indeed... weird.

cyrilleberger:
I will do the merging when #176 is in master, the changes are too complicated for me to do it without proper testing (and probably not before I come back from vacation, so if someone needs the changes, feel free to do the merge).

nuclearsandwich:
@cyrilleberger now that #176 has merged are you able to revisit this PR?

cyrilleberger:
Seems to work and all tests passes (tested with dashing)

cyrilleberger:
How to test is a fair valid point, I have a service call that looks like that (in ROS1 and ROS2):

---
float64[6] transform

I suspect there are no "standard" ROS services that have a fixed size array otherwise you would have implemented that before... If you want some automatic testing, I guess we would need to add such a service call somewhere, I could setup a ros1 and a ros2 package that have just that service definition, if needed.

ros1_bridge/__init__.py
```diff
@@ -583,24 +583,28 @@ def determine_common_services(ros1_srvs, ros2_srvs, mapping_rules):
...
-                ros2_type = str(ros2_fields[direction][i].type)
+                ros2_type = ros2_fields[direction][i].type
                 ros1_name = ros1_field[1]
                 ros2_name = ros2_fields[direction][i].name
-                if ros1_type != ros2_type or ros1_name != ros2_name:
+                if ros1_type != str(ros2_type) or ros1_name != ros2_name:
                     match = False
                     break
+                ros2_cpptype = ros2_type.pkg_name + \
+                    '::msg::' + ros2_type.type if ros2_type.pkg_name else ''
```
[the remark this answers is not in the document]
cyrilleberger: Isn't that handled by "if ros2_type.pkg_name" ? I would have to add that to the test service definition.

cyrilleberger:
> Atm the patch only adds an `assert` statement to check the upper bound of sequences when they come from ROS 1 (where they are unbounded). Is that the scope of the PR? If yes, please update the title to reflect the actual content.

The reason for that (if I remember correctly) ROS1 make a distinction in the generated c++ between dynamic (of type std::vector) and static (of type boost::array?) while ROS2 uses std::vector for both... Now that I say that, it would probably make more sense to assert only in the ROS2 case rather than ROS1.

dirk-thomas:
Closing due to inactivity. Please feel free to comment on the closed ticket and it can be reopened.

cyrilleberger:
Example of srvs https://github.com/cyrilleberger/fixed_size_array_msgs

PR #177, opened 2019-03-21
Fix parameter bridge for topic if ros1 and ros2 type have a different name

cyrilleberger:
I have set ros1 type to "" since in create_bidirectional_bridge it can find out which type to use for ros1 when only specifying ros2 type.

An alternative fix would be to allow to specify different type for ros1 and ros2 in the parameters.

### DIFF
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -73,7 +73,7 @@ int main(int argc, char * argv[])
 
       try {
         ros1_bridge::BridgeHandles handles = ros1_bridge::create_bidirectional_bridge(
-          ros1_node, ros2_node, type_name, type_name, topic_name, queue_size);
+          ros1_node, ros2_node, "", type_name, topic_name, queue_size);
         all_handles.push_back(handles);
       } catch (std::runtime_error & e) {
         fprintf(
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The message printed above provides misleading information now. Please update that accordingly.

cyrilleberger:
Do you mean the commit message?

dirk-thomas:
Sorry, I though I commented on a specific line. Without that context its not obvious. I was referring to the message printed a few lines above the change: line 69-72.

cyrilleberger:
Is there anything else I need to change?

dirk-thomas:
> Is there anything else I need to change?

Please comment after pushing additional commits to your PRs. Otherwise nobody gets a notification and it is unlikely someone will review the changes.

The change looks good to me. Thanks for the improvement.

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/215/)

dirk-thomas:
> [Build Status](https://ci.ros2.org/job/ci_packaging_linux/215/)

It seems like your fork is outdated which failed the CI build.

Testing a rebased version of your fork - the result looks good: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/216/)

PR #183, opened 2019-04-12
Interface specific compilation units

dirk-thomas:
In short: the PR trades a longer build time for a lower memory consumption during the build :balance_scale:

Resolves ros2/build_cop#106.

Before the patch the generated code produced one `.cpp` file per package. The memory consumption of the compiler therefore scales with the number of messages per package and the template specialization stuff is not light on memory. For packages like `geometry_msgs`, `sensor_msgs` or `std_msgs` which each contain ~30 messages the memory consumption of a single compilation unit could easily reach 2.5 GB. Building the bridge with N threads where N is by default the number of CPU cores could therefore easily exhaust the available memory.

With the patch the generated code produces one `.cpp` file per interface file (message, service). Therefore the memory consumption for a single compilation unit drops to less than 0.5 GB. The trade off is that instead of generating X `.cpp` files (where X is the number of interface packages it now generates Y `.cpp` files (where Y is the number of interface files which is easily an order of magnitude higher). So while the peak memory consumption drops significantly the compilation time will increase. On my system with 8 cores it took ~3min 20s before and ~7min 45s afterwards.

[Build Status](https://ci.ros2.org/job/ci_packaging_linux/186/)

The two commits should not be squashed to keep the first commit which just duplicates one of the templates separate to ease tracking the history.

### DIFF
--- a/resource/interface_factories.cpp.em
+++ b/resource/interface_factories.cpp.em
@@ -1,0 +1,295 @@
+// generated from ros1_bridge/resource/pkg_factories.cpp.em
+
+@###############################################
+@#
+@# Factory template specializations based on
+@# message types of a single ROS 2 package
+@#
+@# EmPy template for generating <pkgname>_factories.cpp
+@#
+@###############################################
+@# Start of Template
+@#
+@# Context:
+@#  - ros2_package_name (str)
+@#    The ROS 2 package name of this file
+@#  - mappings (list of ros1_bridge.Mapping)
+@#    Mapping between messages as well as their fields
+@###############################################
+@
+@{
+from ros1_bridge import camel_case_to_lower_case_underscore
+}@
+#include "rclcpp/rclcpp.hpp"
+#include "@(ros2_package_name)_factories.hpp"
+
+// include builtin interfaces
+#include <ros1_bridge/convert_builtin_interfaces.hpp>
+
+// include ROS 1 services
+@[for service in services]@
+#include <@(service["ros1_package"])/@(service["ros1_name"]).h>
+@[end for]@
+
+// include ROS 2 services
+@[for service in services]@
+#include <@(service["ros2_package"])/srv/@(camel_case_to_lower_case_underscore(service["ros2_name"])).hpp>
+@[end for]@
+
+namespace ros1_bridge
+{
+
+std::shared_ptr<FactoryInterface>
+get_factory_@(ros2_package_name)(const std::string & ros1_type_name, const std::string & ros2_type_name)
+{
+@[if not mappings]@
+  (void)ros1_type_name;
+  (void)ros2_type_name;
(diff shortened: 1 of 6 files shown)


resource/pkg_factories.hpp.em
```diff
@@ -27,12 +27,12 @@ from ros1_bridge import camel_case_to_lower_case_underscore
 #include <ros1_bridge/factory.hpp>
```
clalancette: Nit:
 ```
#include <memory>
#include <string>
```

File: resource/pkg_factories.cpp.em [hunk 2]
```diff
@@ -17,277 +17,43 @@
 @#    Mapping between messages as well as their fields
 @###############################################
 @
```
clalancette: Nit:

```
#include <memory>
#include <string>
```

resource/interface_factories.cpp.em
```diff
@@ -0,0 +1,297 @@
...
+@#  - ros2_package_name (str)
+@#    The ROS 2 package name of this file
+@#  - mappings (list of ros1_bridge.Mapping)
+@#    Mapping between messages as well as their fields
+@###############################################
+@
+@{
+from ros1_bridge import camel_case_to_lower_case_underscore
+}@
+#include "rclcpp/rclcpp.hpp"
```
clalancette: Nit:

```
#include <algorithm>
#include <memory>
#include <string>
```

resource/pkg_factories.hpp.em
dirk-thomas: Added in 8e0eef5d563ee8765fca4b9d0e895e58ec107795.

resource/pkg_factories.cpp.em
dirk-thomas: That isn't necessary of the included corresponding header includes them.

resource/interface_factories.cpp.em
dirk-thomas: Added `algorithm` in a1f6a18acedcec9f23b405a12f001f6eca21ac0a. For the other two same as https://github.com/ros2/ros1_bridge/pull/183#discussion_r275847874.

dirk-thomas:
Squashing the last three commits.

dirk-thomas:
> we should also update the README, as it says it takes a "tremendous amount of memory", which won't be true anymore

Done in cd0ef162c6416da44623f70371cde7b3a40893cf.

dirk-thomas:
> I'm not a huge fan of having the declaration for the `get_factory_<pkg>__msg__<msgname>` methods inside of the `pkg_factories.hpp.em` file.

This has been that way before this patch, it is unrelated to this change and also not really necessary. Therefore I don't plan of changing it as part of this PR.

PR #185, opened 2019-04-18
Fix the generator.

clalancette:
https://github.com/ros2/rosidl/pull/359 removed the backwards
compatibility import of rosidl_adapter.parse from rosidl_parser,
so instead we just directly import rosidl_adapter.parse and
use it as necessary.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


clalancette:
A Linux CI packaging job to show the bridge is fixed: [Build Status](http://ci.ros2.org/job/ci_packaging_linux/187/)

PR #189, opened 2019-05-07
changes to avoid deprecated API's

wjwwood:
Connects to ros2/rclcpp#713

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -58,9 +58,7 @@ public:
     const std::string & topic_name,
     size_t queue_size)
   {
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
-    custom_qos_profile.depth = queue_size;
-    return node->create_publisher<ROS2_T>(topic_name, custom_qos_profile);
+    return node->create_publisher<ROS2_T>(topic_name, rclcpp::QoS(rclcpp::KeepLast(queue_size)));
   }
 
   rclcpp::PublisherBase::SharedPtr
@@ -69,7 +67,9 @@ public:
     const std::string & topic_name,
     const rmw_qos_profile_t & qos_profile)
   {
-    return node->create_publisher<ROS2_T>(topic_name, qos_profile);
+    auto qos = rclcpp::QoS(rclcpp::KeepAll());
+    qos.get_rmw_qos_profile() = qos_profile;
+    return node->create_publisher<ROS2_T>(topic_name, qos);
   }
 
   ros::Subscriber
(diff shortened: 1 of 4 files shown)


include/ros1_bridge/factory.hpp
```diff
@@ -69,7 +67,9 @@ class Factory : public FactoryInterface
     const std::string & topic_name,
     const rmw_qos_profile_t & qos_profile)
   {
-    return node->create_publisher<ROS2_T>(topic_name, qos_profile);
+    auto qos = rclcpp::QoS(rclcpp::KeepAll());
+    qos.get_rmw_qos_profile() = qos_profile;
```
ivanpauno: Why not replacing L70 with:

```cpp
rclcpp::QoS qos(rclcpp::QoSInitialization::from_rmw(qos_profile));
```

I think this method is used nowhere and it's no supposed to be used directly by the user, Maybe it could be just deleted. The method that is used is this other one:
https://github.com/ros2/ros1_bridge/blob/3015b901d0d6d3df4de810fa80757d180fa8297d/include/ros1_bridge/factory.hpp#L55-L62

include/ros1_bridge/factory.hpp
```diff
@@ -120,8 +120,10 @@ class Factory : public FactoryInterface
     callback = std::bind(
       &Factory<ROS1_T, ROS2_T>::ros2_callback, std::placeholders::_1, std::placeholders::_2,
       ros1_pub, ros1_type_name_, ros2_type_name_, node->get_logger(), ros2_pub);
+    auto rclcpp_qos = rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(qos));
```
ivanpauno: Maybe is better:
```cpp
rclcpp::QoS rclcpp_qos(rclcpp::QoSInitialization::from_rmw(qos));
```
And same as above, this method is used nowhere. Maybe, just update the other `create_ros2_subscriber` method to call `node->create_subscription<>()`:
https://github.com/ros2/ros1_bridge/blob/3015b901d0d6d3df4de810fa80757d180fa8297d/include/ros1_bridge/factory.hpp#L97-L108

And delete this one.

include/ros1_bridge/factory.hpp
wjwwood: The reason it is like that is because `rclcpp::QoS qos(rclcpp::QoSInitialization::from_rmw(qos_profile));` only overrides the history and depth settings, it doesn't set any of the others, so I have to do `qos.get_rmw_qos_profile() = qos_profile;` no matter what, and in fact the `QoSInitialization`, in this case is completely unused.

Really, this function signature should probably be deprecated as well, since it takes the `rmw_qos_profile_t`, which would mean that this weird use of the `rclcpp::QoS` api could be removed next version. But that's part of a general need in these pr's to go and find secondary API's which use the `rmw_qos_profile_t` and deprecate those too, like in message filters or in services/actions, etc...

Maybe it could be removed? I was just updating every where I could find.

include/ros1_bridge/factory.hpp
wjwwood: > `rclcpp::QoS rclcpp_qos(rclcpp::QoSInitialization::from_rmw(qos));`

Yeah, I could do that, but I'll probably pass on it unless I need to touch this repository again, since it's just stylistic, the compiler should produce the same thing.

PR #192, opened 2019-05-09
fix build by passing options

dirk-thomas:
Follow up of #189 to make the build pass again.

[Build Status](https://ci.ros2.org/job/ci_packaging_linux/201/)


dirk-thomas:
I will go ahead and merge the PR even though CI is still running. It built locally with the patch which it didn't without - so it should be a strict improvement. That allows us to move forward and iterate on other PRs in this repo...

PR #191, opened 2019-05-08
Adds interface type to ROS2 message type name

hidmic:
Fixes https://github.com/ros2/build_cop/issues/192. This pull request fixes ros1 bridge mapping factory template to take into account the interface type in the message type name. 

* CI packaging Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/200/)
* Packaging Linux [Build Status](http://ci.ros2.org/job/packaging_linux/1459/)

Another one to be added to the list in https://github.com/ros2/ros2/issues/677.

PR #194, opened 2019-05-15
Use interface kind names properly in ROS2 interface type names.

hidmic:
Follow up after 816b667de32f6a5ab45a7cfc22b931f3d2ac1749, introduced by #191. Prefixes (i.e. `msg`, `srv`) were still missing in a few other places.

* CI packaging Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/205/)

I managed to reproduce latest test failures on https://github.com/ros2/build_cop/issues/192 locally and all appear to be fixed by this patch.


hidmic:
Alright, going in!

PR #196, opened 2019-05-17
Disable rosout logging for Fast-RTPS tests

hidmic:
Closes https://github.com/ros2/build_cop/issues/192 (again). Incidentally, I came across some typos in the CMakeLists.txt affecting the tests (!) that got past review/CI before. Those are fixed in their own separate commit.

* CI packaging Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/212/)


dirk-thomas:
While this patch fixes the test it doesn't address the same problem for users running the bridge, right?

hidmic:
Correct. Are you thinking of forcing the configuration behind the scenes?

dirk-thomas:
Yes, the bridge node would always apply that flag - even if not specified on the command line. As least that is the only way I see this would work for users...

hidmic:
Superseded by https://github.com/ros2/ros1_bridge/pull/197.

PR #204, opened 2019-06-20
Use %zu print format for size_t

emersonknapp:
One more brick in the wall of https://github.com/ros2/ros2/issues/721

Fixes warnings seen here https://ci.ros2.org/job/ci_packaging_linux-armhf/4/warnings23Result/package.1964502061/

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -702,7 +702,7 @@ int main(int argc, char * argv[])
         }
 
         if (output_topic_introspection) {
-          printf("  ROS 2: %s (%s) [%ld pubs, %ld subs]\n",
+          printf("  ROS 2: %s (%s) [%zu pubs, %zu subs]\n",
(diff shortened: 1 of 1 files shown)


emersonknapp:
Linux-armhf Packaging: [Build Status](https://ci.ros2.org/job/ci_packaging_linux-armhf/6/)

emersonknapp:
@clalancette what CI should I run for this PR? I'm running ci_packaging_linux-armhf since the packaging jobs are the only ones I know of that build the ros1_bridge. I expect that one to come out green now since this should be the last warning.

clalancette:
> @clalancette what CI should I run for this PR? I'm running ci_packaging_linux-armhf since the packaging jobs are the only ones I know of that build the ros1_bridge. I expect that one to come out green now since this should be the last warning.

Yes, that's the correct one; as you say, it's the only one that builds the bridge.

Issue #205, opened 2019-06-21
ros1_bridge dynamic_bridge missing libroscpp.so


dario-pedro:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04 running docker ROS2 nightly
- Installation type:
  - docker ROS2 nightly

When I run `docker ROS2 nightly` I get /opt/ros/dashing/lib/ros1_bridge/dynamic_bridge: error while loading shared libraries: libroscpp.so: cannot open shared object file: No such file or directory

Any idea how to solve it ?

dirk-thomas:
Have you sourced ROS 1?

kk2491:
@dirk-thomas @dario-pedro I was getting the same error. So I sourced both ```dashing``` and ```melodic```, now I am getting the below error. Could you please let me know what the issue is.
```
system@d1dbb9f4c6b7:/module$ ros2 run ros1_bridge dynamic_bridge 
Failed to load entry point 'info': /opt/ros/dashing/lib/librosbag2_storage.so: undefined symbol: _ZNK12class_loader11ClassLoader15isLibraryLoadedEv
The C extension '/opt/ros/dashing/lib/python3.6/site-packages/rosbag2_transport/_rosbag2_transport_py.cpython-36m-x86_64-linux-gnu.so' failed to be imported while being present on the system. Please refer to 'https://index.ros.org/doc/ros2/Troubleshooting/#import-failing-even-with-library-present-on-the-system' for possible solutions
Segmentation fault
systeml@d1dbb9f4c6b7:/module$ 
```

dirk-thomas:
> I sourced both `dashing` and `melodic`

You need to source them in the other order - ROS 1 first, then ROS 2 - as described in the instructions.

dirk-thomas:
Assuming the previous comments have resolved the issue I will go ahead an close this ticket. Please feel free to continue commenting and if necessary the ticket can be reopened.

potentialdiffer:
I am running into a similar problem. (Using Debian 9, Kinetic and Dashing from source) What I've done so far: 

- rebuild ROS1 and ROS2 workspace in "clean" terminal
- compiled `ros1_bridge` with first sourcing ROS1 and then ROS2 workspace 

Building `ros1_bridge` succeeds, however running the bridge results in the error, which was previously mentioned:

```
$ source /home/xxx/ros1_kinetic/install/setup.bash
$ source /home/xxx/ros2_dashing/install/setup.bash
[...]

PR #211, opened 2019-08-11
Clearer instructions for example ros1->ros2

jlblancoc:
Related to #205. I also found the instructions unclear about which ROS versions one has to source to run the bridge, so the proposal is to make it clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ You will get errors from most tools if they have both workspaces in their enviro
 First we start a ROS 1 `roscore`:
 
 ```
-# Shell A:
+# Shell A (ROS 1 only):
 . /opt/ros/melodic/setup.bash
 # Or, on OSX, something like:
 # . ~/ros_catkin_ws/install_isolated/setup.bash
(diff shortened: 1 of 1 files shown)


jlblancoc:
Agreed DCO after opening the PR... probably you can re-run the check? if not, feel free of closing and re-opening a new one. Cheers

README.md
```diff
@@ -107,7 +107,7 @@ You will get errors from most tools if they have both workspaces in their enviro
 First we start a ROS 1 `roscore`:
 
 ```
-# Shell A:
+# Shell A (ROS1 only):
```
dirk-thomas: Please use a space between `ROS` and the major version number.

Same below multiple times.

dirk-thomas:
> Agreed DCO after opening the PR... probably you can re-run the check?

Actually you have to update the commit message to contain the line `Signed-off-by: ...` (see https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#pull-requests). You can fix the existing commit with `git commit --amend -s` and then force push that change to your branch: `git push -f`. The DCO check will pass.

jlblancoc:
Oh, sure! Didn't quite understood the purpose of the DCO app.
Done now. Thanks @dirk-thomas .

dirk-thomas:
Thanks for the improvement.

PR #214, opened 2019-08-23
Support for dynamic whitelisting bridge using regular expressions

dhananjaysathe:
Adds a new flavor of the bridge that supports adding a list of regular expressions as parameters to selectively and dynamically forward bridge topics/services.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,7 +257,7 @@ macro(targets)
   add_launch_test(
      "${CMAKE_CURRENT_BINARY_DIR}/test_services_across_dynamic_bridge${target_suffix}_$<CONFIG>.py"
      TARGET test_services_across_dynamic_bridge${target_suffix}
-     ENV RMW_IMPLEMENTATION=${rmw_implementation}
+     ENV RMW_IMPLEMENTATION=${rmw_implementation})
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Thank you for working on this feature. Why should the whitelisting be implemented in a separate executable rather than adding these options to the existing dynamic bridge? If none are specified the behavior would stay the same.

Also the added test might be good to separate into its own pull request to cover the existing dynamic bridge.

dhananjaysathe:
Hey @dirk-thomas the bridge is a new executable called "dynamic_whitelist_bridge"  its similar sounding or named what led to the confusion i think .

dirk-thomas:
My previous comment was asking *why* the whitelisting should be implemented in a separate executable. Why not add the feature to the *eixsting* executable `dynamic_bridge`?

dhananjaysathe:
Ah i see let me chase this up next week

dirk-thomas:
@dhananjaysathe Any update on this?

claireyywang:
@dhananjaysathe friendly ping.

dirk-thomas:
Closing due to no response. Please feel free to comment on the closed ticket and it can be reopened.

PR #224, opened 2019-10-22
fix bridging builtin_interfaces Duration and Time

dirk-thomas:
Based on https://answers.ros.org/question/335602/ros1_bridge-for-builtin-interfaces-results-in-an-error/


dirk-thomas:
[Build Status](https://ci.ros2.org/view/All/job/ci_packaging_linux/270/)

PR #230, opened 2019-12-05
[backport dashing] fix showing duplicate keys in --print-pairs

dirk-thomas:
Backport of #225.

### DIFF
--- a/include/ros1_bridge/bridge.hpp
+++ b/include/ros1_bridge/bridge.hpp
@@ -58,10 +58,10 @@ get_2to1_mapping(
   const std::string & ros2_type_name,
   std::string & ros1_type_name);
 
-std::map<std::string, std::string>
+std::multimap<std::string, std::string>
 get_all_message_mappings_2to1();
 
-std::map<std::string, std::string>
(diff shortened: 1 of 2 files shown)


include/ros1_bridge/bridge.hpp
```diff
@@ -58,10 +58,10 @@ get_2to1_mapping(
   const std::string & ros2_type_name,
   std::string & ros1_type_name);
 
-std::map<std::string, std::string>
+std::multimap<std::string, std::string>
```
dirk-thomas: This API change is necessary to get the necessary information to fix `--print-pairs`. Either we are ok with that API breakage (and notify users in the sync email about it) or we should close this backport and leave the `--print-pairs` in the broken state in Dashing.

wjwwood: I would vote for closing to avoid API changes, or add a new function (different name) only in dashing and update our code to use that one instead.

dirk-thomas:
Ok, with 3 x :-1: I will close this backport. Thanks for the feedback.

PR #233, opened 2020-01-09
Drop workaround for Fast-RTPS typesupport issue 

hidmic:
Precisely what the title says. To be tested and merged along with https://github.com/ros2/rmw_fastrtps/pull/342.

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -459,22 +459,7 @@ int main(int argc, char * argv[])
   }
 
   // ROS 2 node
-
-  // TODO(hidmic): remove when Fast-RTPS supports registering multiple
-  //               typesupports for the same topic in the same process.
-  //               See https://github.com/ros2/rmw_fastrtps/issues/265.
-  std::vector<const char *> args(argv, argv + argc);
-
-  const char * rmw_implementation = "";
-  const char * error = rcutils_get_env("RMW_IMPLEMENTATION", &rmw_implementation);
-  if (NULL != error) {
-    throw std::runtime_error(error);
-  }
-  if (0 == strcmp(rmw_implementation, "") || NULL != strstr(rmw_implementation, "fastrtps")) {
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `ros1_bridge`:
* Packaging Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/280/)

dirk-thomas:
Depending on the answer of this question https://github.com/ros2/rmw_fastrtps/pull/342#issuecomment-574438327 this change should be considered for backporting.

hidmic:
Hmm, `ros1_bridge` seems to exhibit segfaults when using `rmw_fastrtps_dynamic_cpp` only. I'll investigate.

hidmic:
This is blocked until https://github.com/eclipse-cyclonedds/cyclonedds/issues/339 and https://github.com/ros2/rmw_fastrtps/issues/265 are finally addressed. Which leaves us in the same place we were at when this workaround was first introduced. I pinged eProsima and ADLINK folks, but we may as well close this until those fixes land.

MiguelCompany:
> This is blocked until [eclipse-cyclonedds/cyclonedds#339](https://github.com/eclipse-cyclonedds/cyclonedds/issues/339) and [ros2/rmw_fastrtps#265](https://github.com/ros2/rmw_fastrtps/issues/265)

@hidmic I then suggest to change the title of this PR.

hidmic:
> I then suggest to change the title of this PR.

Fair enough :)

dirk-thomas:
I changed the title back since the patch knows nothing about CycloneDDS but only checks for FastRTPS.

hidmic:
Oh, that's true. Silly me. I wonder why no one has run into this issue using CycloneDDS.

dirk-thomas:
> I wonder why no one has run into this issue using CycloneDDS.

The test in the packaging job is failing with CycloneDDS for that reason.

PR #234, opened 2020-01-10
fix service with custom mapped message field

dirk-thomas:
Based on this bug report: https://answers.ros.org/question/340860/ros1_bridge-fails-to-generate-service-when-custom-message-from-the-same-package-is-used/

Packaging build: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/281/) (unrelated warnings: ros2/rclcpp#962)


dirk-thomas:
Synced change to `eloquent` branch.

PR #236, opened 2020-01-15
fix building test when ROS 1 diagnostic_msgs is isolated from roscpp

dirk-thomas:


dirk-thomas:
Synced the `eloquent` branch with `master`.

PR #242, opened 2020-03-04
add support for fixed size array in service calls

cyrilleberger:
Same as #175 (couldn't reopen), with example of srvs https://github.com/cyrilleberger/fixed_size_array_msgs .


ros1_bridge/__init__.py
```diff
@@ -596,27 +596,34 @@ def determine_common_services(
...
+                else:
+                    upper_bound_array = None
+                ros2_cpptype = ros2_type.pkg_name + \
+                    '::msg::' + ros2_type.type if ros2_type.pkg_name else ''
                 output[direction].append({
-                    'basic': False if '/' in ros1_type else True,
-                    'array': True if '[]' in ros1_type else False,
+                    'basic': not ros2_type.pkg_name,
+                    'array': ros2_type.is_array,
+                    'upper_bound_array': upper_bound_array,
```
dirk-thomas: I would rather not introduce a new term "upper_bound_array" where it is unclear what the semantic is. Instead the existing values of `array_size` and `is_upper_bound` should be passed into the template.

ros1_bridge/__init__.py
```diff
@@ -596,27 +596,34 @@ def determine_common_services(
...
-                    if (ros1_type, ros2_type) not in message_string_pairs:
+                    if (ros1_type, str(ros2_type)) not in message_string_pairs:
                         match = False
                         break
+                if not ros2_type.array_size and not ros2_type.is_upper_bound:
+                    upper_bound_array = ros2_type.array_size if ros2_type.is_upper_bound else -1
+                else:
+                    upper_bound_array = None
+                ros2_cpptype = ros2_type.pkg_name + \
+                    '::msg::' + ros2_type.type if ros2_type.pkg_name else ''
```
dirk-thomas: I don't understand the `else` case here. How can the C++ type name be an empty string?

cyrilleberger:
Is there anything else I need to fix? It would be nice if it was finally merged

ros1_bridge/__init__.py
```diff
@@ -596,27 +596,37 @@ def determine_common_services(
...
-                if ros1_type != ros2_type or ros1_name != ros2_name:
+                if ros1_type != str(ros2_type) or ros1_name != ros2_name:
                     # if the message types have a custom mapping their names
                     # might not be equal, therefore check the message pairs
-                    if (ros1_type, ros2_type) not in message_string_pairs:
+                    if (ros1_type, str(ros2_type)) not in message_string_pairs:
                         match = False
                         break
+                if not ros2_type.array_size and not ros2_type.is_upper_bound:
+                    array_size = ros2_type.array_size if ros2_type.is_upper_bound else -1
```
dirk-thomas, line 609: I don't understand what the purpose of these two lines is. After the condition in line 608 `ros2_type.is_upper_bound` must be `False` in order to enter the `if` block. So in this line that means `array_size` is always set to `-1` - never to `ros2_type.array_size` (which based on the condition in line 608 would need to be `None` anyway).

ros1_bridge/__init__.py
```diff
@@ -596,27 +596,37 @@ def determine_common_services(
...
+                    ros2_cpptype = ros2_type.pkg_name + "::msg::" + ros2_type.type
+                else:
+                    ros2_cpptype = str(ros2_type)
                 output[direction].append({
-                    'basic': False if '/' in ros1_type else True,
-                    'array': True if '[]' in ros1_type else False,
+                    'basic': not ros2_type.pkg_name,
+                    'array': ros2_type.is_array,
+                    'is_upper_bound': ros2_type.is_upper_bound,
+                    'array_size': array_size,
```
dirk-thomas, line 620: Similar to my previous comment I suggest to avoid any kind of indirection and changing the semantic is existing names. Why not just pass `ros2_type.array_size` here?

Based on the above comment `array_size` is either `-1` or `None` here which doesn't seem to make much sense.

resource/interface_factories.cpp.em
```diff
@@ -277,7 +277,15 @@ void ServiceFactory<
 ) {
 @[      for field in service["fields"][type.lower()]]@
 @[        if field["array"]]@
+@[          if field["is_upper_bound"]]@
+  // ensure array size
+@[            if field["array_size"] > 0 and frm == "1"]@
+  // check boundary
+                assert(req@(frm).@(field["ros1"]["name"]).size() <= @(field["array_size"]));
```
dirk-thomas, line 284: Based on the other comments I don't think this `assert` is ever being generated.

dirk-thomas:
@cyrilleberger Friendly ping.

cyrilleberger:
I am sorry I have spent too much time and effort that I can allocate on this, I have instead reported it as bug #283 and changed the single service call I have with a fixed array to a dynamic one.

PR #245, opened 2020-03-10
Ignore actionlib_msgs deprecation warnings

jacobperron:
Fixes https://github.com/ros2/build_farmer/issues/267
Depends on https://github.com/ament/ament_cmake/pull/223

Note, this change will ignore *all* deprecation warnings produced by calling `find_package(actionlib_msgs)`. I'm not sure of a nice way in CMake to ignore a single message coming from a command, but I think this solution is reasonable since I don't expect other deprecation warnings to come from finding `actionlib_msgs`. Alternatively, we could introduce a new variable in `ament_cmake` to toggle the deprecation warning, e.g. `actionlib_msgs_IGNORE_DEPRECATED`, but it doesn't seem very elegant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,17 +90,11 @@ list(APPEND generated_files "${generated_path}/get_mappings.cpp")
 
 # generate per interface compilation units to keep the memory usage low
 ament_index_get_resources(ros2_interface_packages "rosidl_interfaces")
+# actionlib_msgs is deprecated, but we will quiet the warning until the bridge has support for
+# ROS actions: https://github.com/ros2/design/issues/195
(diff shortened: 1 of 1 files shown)


dirk-thomas:
If the referenced PR is being updated with this comment (https://github.com/ament/ament_cmake/pull/223#issuecomment-597364775) only that specific deprecation message can be silenced.

jacobperron:
I've updated both PRs to allow for alternative approach, which quiets the single warning we care to ignore.

jacobperron:
* Linux packaging [Build Status](https://ci.ros2.org/job/ci_packaging_linux/301/)

PR #246, opened 2020-03-11
👩‍🌾 [Proposal] No warning for missing ROS 1

chapulina:
This warning keeps coming up on [ROS 2's CI](http://build.ros2.org/view/Fci/job/Fci__nightly-release_ubuntu_focal_amd64/44/warnings2Result/), but if it's not something that needs to be fixed, I'd propose to turn into an informational message instead of a warning (which prompts action).

I'm not sure if this has been discussed in the past, and I can see an argument for keeping it as it is. I was just wondering what people thought.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ endif()
 
 find_ros1_package(roscpp)
 if(NOT ros1_roscpp_FOUND)
-  message(WARNING "Failed to find ROS 1 roscpp, skipping...")
+  message(STATUS "Failed to find ROS 1 roscpp, skipping...")
   # call ament_package() to avoid ament_tools treating this as a plain CMake pkg
   ament_package(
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think this should be merged. The warning is there for a reason: if a user tries to build this package and ROS 1 can't be found there needs to be a warning.

Instead the CI builds showing this warning should be updated to not build `ros1_bridge` - or setup a ROS 1 environment if that is desired.

chapulina:
Got it, makes sense.

It sounds like this package can't be used without ROS1. Then in this case, shouldn't this be an error instead?

dirk-thomas:
Since the repo is part of the `ros2.repos` file which we recommend in the ROS 2 build-from-source instructions I don't think a `FATAL_ERROR` which will essentially fail the build is an option.

PR #176, opened 2019-03-21
add services bridging to parameter_bridge

cyrilleberger:
making possible to define services using the parameter_bridge, it is done very similarly to topics:
It needs to parameter 'services_1_to_2' (for calling ROS2 services from ROS1) and 'services_2_to_1 (for calling ROS1 from ROS2), I tried to follow the naming convention of the dynamic_bridge. The parameters should look like:
[
{service: /name/of/service package: name_of_package_with_service, type: type_of_message},
...
]
For calling ROS1 add_two_ints from ROS2:
[
{service: /add_two_ints package: roscpp_tutorials, type: TwoInts}
]

PR #249, opened 2020-03-31
[parameter_bridge] add srv to the type to keep yaml description consistent

mikaelarguedas:
Similar to https://github.com/ros2/ros1_bridge/pull/194

Before this PR the syntax for bridging services in both directions was the following:
<details>

```yaml
---
# service requests going from 2 to 1
# server on ROS1 side, client on ROS2 side
services_2_to_1:
  -
    service: /add_two_ints
    package: roscpp_tutorials
    type: TwoInts
# service requests going from 1 to 2
# server on ROS2 side, client on ROS1 side
services_1_to_2:
  -
    service: /add_two_ints
    package: example_interfaces
    type: srv/AddTwoInts
```

</details>

Having to specify the type as `srv/AddTwoInts` seems uncommon and counter intuitive

<details><summary>After this PR</summary>

```yaml
---
# service requests going from 2 to 1
# server on ROS1 side, client on ROS2 side
services_2_to_1:
  -
    service: /add_two_ints
    package: roscpp_tutorials
    type: TwoInts
# service requests going from 1 to 2
# server on ROS2 side, client on ROS1 side
services_1_to_2:
  -
    service: /add_two_ints
    package: example_interfaces
    type: AddTwoInts
```

</details>

---

This is still breaking pattern from what is used to specify topic bridging:

```yaml
topics: 
  -
    topic: /clock
    type: rosgraph_msgs/msg/Clock
```

Not sure what the reason was in the first place for splitting the package and the type name only for services, but this PR keeps that behavior.

### DIFF
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -109,7 +109,7 @@ int main(int argc, char * argv[])
     for (size_t i = 0; i < static_cast<size_t>(services_1_to_2.size()); ++i) {
       std::string service_name = static_cast<std::string>(services_1_to_2[i]["service"]);
       std::string package_name = static_cast<std::string>(services_1_to_2[i]["package"]);
-      std::string type_name = static_cast<std::string>(services_1_to_2[i]["type"]);
+      std::string type_name = "srv/" + static_cast<std::string>(services_1_to_2[i]["type"]);
       printf(
         "Trying to create bridge for ROS 2 service '%s' "
         "with package '%s' and type '%s'\n",


hidmic:
Hmm, I'd be inclined towards keeping it consistent with how message bridges are described (i.e. using just `type`) instead of silently adding prefixes. Also, to not break existing code that's using the current syntax, accepting it still and printing a deprecation warning would be nice. WDYT? @dirk-thomas thoughts?

hidmic:
@mikaelarguedas friendly ping.

mikaelarguedas:
This is really just a fix because the "srv" was silently added in https://github.com/ros2/ros1_bridge/pull/194/files#diff-72198d9d0a32d2518cd7a4cedb8d26caR89 forcing people to add it by hand to their yaml file.

So it doesn't strike me as a case where this needs to preserve backward compatibility.
If this is a must-have to get this merged I can try to find the bandwidth to look at it more closely.

Regarding describing services similarly to topics, it would be great. This could result in a significant refactor that I won't be able to work on by the foxy freeze so I'd prefer to target a minimum convenience change.

mikaelarguedas:
@hidmic how do you feel about this ? is there anything that needs to be changed for this to get into Foxy ?

hidmic:
Ideally, I'd (personally) prefer something closer to https://github.com/ros2/ros1_bridge/pull/249#issuecomment-611006124. But we can move forward with this as an incremental improvement. Do you have the bits to run CI?

hidmic:
CI up to `ros1_bridge`:
* Packaging Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/351/)

dirk-thomas:
Support for services was added in #176 to the parameter bridge. That state has been released into Eloquent so I think it is desirable to keep backward compatibility with potentially existing parameters.

Please see #263 which deprecates the `package` key - but should keep it working. When no `package` key is defined the `type` values has the same pattern as for topics.

PR #252, opened 2020-04-06
Do not generate files for interfaces in rmw_dds_common

ivanpauno:
Fixes https://github.com/ros2/rmw_dds_common/issues/8.

There's no reason to generate files for the interfaces in `rmw_dds_common`, so I propose to skip them.

Another way to skip the package, is to [avoid registering it](https://github.com/ros2/rosidl/blob/4d837670d5758a3d45b8c86f3d867f44edd1c221/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L252) as a `rosidl_interfaces` resource.
I can add an argument to [rosidl_generate_interfaces](https://github.com/ros2/rosidl/blob/4d837670d5758a3d45b8c86f3d867f44edd1c221/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L53) to do that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,9 @@ list(APPEND generated_files "${generated_path}/get_mappings.cpp")
 
 # generate per interface compilation units to keep the memory usage low
 ament_index_get_resources(ros2_interface_packages "rosidl_interfaces")
+# Do not generate factories for interfaces in rmw_dds_common
+list(REMOVE_ITEM ros2_interface_packages rmw_dds_common)
+
 # actionlib_msgs is deprecated, but we will quiet the warning until the bridge has support for
 # ROS actions: https://github.com/ros2/design/issues/195
 set(actionlib_msgs_DEPRECATED_QUIET TRUE)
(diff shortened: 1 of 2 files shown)


ivanpauno:
* Packaging Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/313/)

dirk-thomas:
This package should not need to know about `rmw_dds_common` so this should *not* be merged 

I don't know enough about how the interface in that package are needed to recommend an approach.

ivanpauno:
> Another way to skip the package, is to avoid registering it as a rosidl_interfaces resource.
I can add an argument to rosidl_generate_interfaces to do that.

Does this sound reasonable?

dirk-thomas:
I don't understand why they must be installed but can't be registered at the index.

ivanpauno:
> I don't understand why they must be installed but can't be registered at the index.

They must be installed because `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp` use them.

They don't need to be registered at the index because they are not intended to be an interface used by the end user.

ivanpauno:
See https://github.com/ros2/rmw_dds_common/pull/11 and https://github.com/ros2/rosidl/pull/454.

dirk-thomas:
> They don't need to be registered at the index because they are not intended to be an interface used by the end user.

The question remains: what is the problem for them being registered at the index?

Issue #257, opened 2020-05-08
Explicitly mapping packages with the same package name and service names causes a redefinition error on build


suddrey-qut:
Operating System:
Description: Ubuntu 18.04.4 LTS

Installation type:
eloquent binaries, ros1_bridge source

Version or commit hash:
On eloquent branch of ros1_bridge:
± git rev-parse HEAD
60f903cb159ad4ef59b9b09badf71caa601cba04

DDS implementation:
Fast-RTPS

Client library (if applicable):
N/A

#### Steps to reproduce issue
Following the instructions found on the [ros1_bridge docs](https://industrial-training-master.readthedocs.io/en/melodic/_source/session7/ROS1-ROS2-bridge.html).

Clone the ROS1 version of the messages into a ROS1 workspace and build it
```
git clone git@github.com:RoboticVisionOrg/rv_msgs.git -b ros1_bridge_example
```
Clone the ROS2 version of the messages into a ROS2 workspace and build it
```
git clone git@github.com:RoboticVisionOrg/rv_msgs.git -b eloquent-devel
```

In a new terminal, ensuring no work spaces are sourced clone the eloquent branch of ros1_bridge into a new empty workspace
```
git clone https://github.com/ros2/ros1_bridge.git -b eloquent
```

- source the ROS1 workspace
- source the ROS2 workspace
- build the ros1_bridge 

```
colcon build --packages-select ros1_bridge --cmake-force-configure --cmake-args -DBUILD_TESTING=FALSE
```

#### Expected behavior
Bridge compiles without errors

#### Actual behavior
generate_services adds a pair for both the implicit match derived from rospack and the explicit match provided in the mapping.yaml file causing gcc to throw redefinition errors.

```
[...]

PR #259, opened 2020-05-08
When generating service mappings cast pair to set to avoid duplicate pairs

suddrey-qut:
Fixes issue #257

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -564,15 +564,21 @@ def determine_common_services(
             for ros2_srv in ros2_srvs:
                 if rule.ros1_package_name == ros1_srv.package_name and \
                    rule.ros2_package_name == ros2_srv.package_name:
+
+                    pair = (ros1_srv, ros2_srv)
+
+                    if pair in pairs:
+                        continue
+
                     if rule.ros1_service_name is None and rule.ros2_service_name is None:
                         if ros1_srv.message_name == ros2_srv.message_name:
-                            pairs.append((ros1_srv, ros2_srv))
+                            pairs.append(pair)
                     else:
                         if (
                             rule.ros1_service_name == ros1_srv.message_name and
                             rule.ros2_service_name == ros2_srv.message_name
                         ):
-                            pairs.append((ros1_srv, ros2_srv))
+                            pairs.append(pair)
 
     for pair in pairs:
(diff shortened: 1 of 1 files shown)


ros1_bridge/__init__.py
```diff
@@ -574,7 +574,7 @@ def determine_common_services(
                         ):
                             pairs.append((ros1_srv, ros2_srv))
 
-    for pair in pairs:
+    for pair in set(pairs):
```
peterpena, line 577: In `determine_message_pairs`, the pairs are added beforehand and then only new pairs are added to the set; otherwise the for loop continues. A similar construct is done in `determine_common_services`, but all the pairs are added and then, in your fix, the pairs object is passed to `set` to give you a unique set of pairs. 

My question is can we have a solution that follows a similar construct in `determine_message_pairs` where both functions follow a similar implementation? Both functions have similar functionality, and therefore, should have similar implementation. It will help with readibility.

CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 Changelog for package ros1_bridge
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+0.8.2 (2020-01-17)
+------------------
+* fix building test when ROS 1 diagnostic_msgs is isolated from roscpp (`#236 <https://github.com/ros2/ros1_bridge/issues/236>`_)
+* fix service with custom mapped message field (`#234 <https://github.com/ros2/ros1_bridge/issues/234>`_)
+* Contributors: Dirk Thomas
+
```
dirk-thomas, line 10: Please revert these unrelated changes.

package.xml
```diff
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>ros1_bridge</name>
-  <version>0.8.1</version>
+  <version>0.8.2</version>
```
dirk-thomas, line 5: Same here.

hidmic:
@suddrey-qut are you still willing to push for this changes?

suddrey-qut:
Hi, yes I will make these changes shortly. Been a bit smashed with work sorry.

suddrey-qut:
Going to recreate this pull request based on master branch.

Issue #38, opened 2016-12-04
Bridges to ROS 2 publishers/subscribers not removed when using Fast RTPS


dhood:
Connects to ros2/rmw_fastrtps#64 - opening this ticket here for visibility.

Bridges aren't always removed when expected (when ROS 2 nodes shut down) if `rmw_fastrtps_cpp` is used as the rmw implementation

Kukanani:
I just confirmed that this is still an issue, at least when the publishing ROS2 node is killed via Ctrl-C.

Karsten1987:
The same goes for count subscriber / count publisher btw

dirk-thomas:
@Kukanani @Karsten1987 have you both used the latest version from the default branches (including FastRTPS and rmw_fastrtps_cpp)?

Kukanani:
Updating appears to have fixed the issue on my machine.

Karsten1987:
bridge works for me. The output of "--show-introspection" shows that the ros2 node correctly disappears.

dirk-thomas:
Based on the latest reports confirming that the current state works I will close this ticket for now. Please feel free to comment here if you still experience the problem and the ticket can be reopened.

PR #260, opened 2020-05-10
Remove outdated information on Fast RTPS bug

Thom747:
https://github.com/ros2/ros1_bridge/issues/38 has been resolved.


dirk-thomas:
@Thom747 Thanks for the patch. Please update your commit to include a *Signed-off-by* line in order to pass the DCO check.

dirk-thomas:
Merging - even without the DCO passing - since the patch only removes a paragraph from the README.

PR #263, opened 2020-05-20
deprecate package key for service parameters, use full type instead

dirk-thomas:
Replaces #249.

CI packaging build: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/354/)

### DIFF
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -129,7 +129,7 @@ int main(int argc, char * argv[])
         "Trying to create bridge for ROS 2 service '%s' with type '%s'\n",
         service_name.c_str(), type_name.c_str());
 
-      size_t index = type_name.find("/");
+      const size_t index = type_name.find("/");
       if (index == std::string::npos) {
         fprintf(
           stderr,
@@ -191,7 +191,7 @@ int main(int argc, char * argv[])
         "Trying to create bridge for ROS 1 service '%s' with type '%s'\n",
         service_name.c_str(), type_name.c_str());
 
-      size_t index = type_name.find("/");
+      const size_t index = type_name.find("/");
       if (index == std::string::npos) {
         fprintf(
           stderr,
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -46,12 +46,16 @@ int main(int argc, char * argv[])
   std::list<ros1_bridge::ServiceBridge2to1> service_bridges_2_to_1;
 
   // bridge all topics listed in a ROS 1 parameter
-  // the parameter needs to be an array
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@mikaelarguedas Can you please give this patch a try. Thanks.

mikaelarguedas:
I won't be behind a computer for the long weekend so I'll be able to test this patch on Monday at the earliest.

src/parameter_bridge.cpp
```diff
@@ -108,15 +112,33 @@ int main(int argc, char * argv[])
...
+        }
+      }
       printf(
-        "Trying to create bridge for ROS 2 service '%s' "
-        "with package '%s' and type '%s'\n",
-        service_name.c_str(), package_name.c_str(), type_name.c_str());
+        "Trying to create bridge for ROS 2 service '%s' with type '%s'\n",
+        service_name.c_str(), type_name.c_str());
 
+      size_t index = type_name.find("/");
```
hidmic, line 132: @dirk-thomas nit^N: can be made `const`

src/parameter_bridge.cpp
```diff
@@ -154,15 +174,34 @@ int main(int argc, char * argv[])
...
+        }
+      }
       printf(
-        "Trying to create bridge for ROS 1 service '%s' "
-        "with package '%s' and type '%s'\n",
-        service_name.c_str(), package_name.c_str(), type_name.c_str());
+        "Trying to create bridge for ROS 1 service '%s' with type '%s'\n",
+        service_name.c_str(), type_name.c_str());
+
+      size_t index = type_name.find("/");
```
hidmic, line 194: @dirk-thomas nit^N: can be made `const`

mikaelarguedas:
Tested his on top of the nightly image and it works fine in both directions :+1:

Parameter file used for testing:
```yaml
---
# service requests going from 2 to 1
# server on ROS1 side, client on ROS2 side
services_2_to_1:
  -
    service: /add_two_ints
    # package: roscpp_tutorials
    type: roscpp_tutorials/TwoInts
# service requests going from 1 to 2
# server on ROS2 side, client on ROS1 side
services_1_to_2:
  -
    service: /add_two_ints
    type: example_interfaces/srv/AddTwoInts

```

src/parameter_bridge.cpp
dirk-thomas: Done in 5ea5a40c23a65cef995caba7d8334ed482d65c8f.

src/parameter_bridge.cpp
dirk-thomas: Done in 5ea5a40c23a65cef995caba7d8334ed482d65c8f.

PR #267, opened 2020-05-22
fix removing obsolete ROS 2 service bridges

### LINKED ISSUE #265: ROS2-side server bridge stays open after the server is destroyed
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Focal

- Installation type:
  - From source 

- Version or commit hash:
  - 5f70a38cf4ad9f11767350052552c260dd5301e9

- DDS implementation: 
  - Fast-RTPS

- Client library:
  - rclcpp and rclpy

#### Steps to reproduce issue

Launch a ROS1 master:
```
# Shell A:
. <ros-install-dir>/setup.bash
roscore -p 11311
```

Launch the `dynamic_bridge`:
```
# Shell B:
. <ros-install-dir>/setup.bash
. <ros2-install-dir>/setup.bash
export ROS_MASTER_URI=http://localhost:11311
ros2 run ros1_bridge dynamic_bridge
```
Launch a server on the ROS2-side:

```
# Shell C:
. <ros2-install-dir>/setup.bash
ros2 run demo_nodes_py add_two_ints_server
```

Now, stop the server in shell C (`Ctrl+C`) and run a server on the ROS1-side:

```
# Shell D:
. <ros-install-dir>/setup.bash
export ROS_MASTER_URI=http://localhost:11311
rosrun rospy_tutorials add_two_ints_server
```

Launch a client on the ROS2 side:

```
# Shell C (continued):
ros2 run demo_nodes_py add_two_ints_client
```

The service won't be found by the client and you'll see the message `service not available, waiting again...`.

In my experience, the `dynamic_bridge` has to be restarted before I can start a service with the same service name.  

You can try going in the other direction: start with a service on ROS1 before trying a service on ROS2 and it should work.  Notice that on the bridge, it says `Created 2 to 1 bridge for service /add_two_ints` when the server is created and then `Removed 2 to 1 bridge for service /add_two_ints` when the server is stopped.  On the ROS2 side, the bridge is created, but the bridge is not removed when the service is stopped.  I have found this to be true for both the C++ and Python examples.

#### Expected behavior
That I'm able to create a server with the same name on either the ROS1-side or ROS2-side, close that server, and then create the server on the other side.

#### Actual behavior
I'm not able to discover the server created on the ROS1-side of the bridge after it has been created and the node has been deleted on the ROS2-side.  It seems that this is because the bridge for the ROS2 server is never removed.

dirk-thomas:
Fixes #265. Requires ros2/rclcpp#1131.

CI builds testing `rclcpp` with only FastRTPS:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10872/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6235/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8856/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10765/)

Packaging build with only FastRTPS: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/359/)

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -695,6 +695,21 @@ int main(int argc, char * argv[])
         }
       }
 
+      // collect available services (not clients)
+      std::set<std::string> service_names;
+      std::vector<std::pair<std::string, std::string>> node_names_and_namespaces =
+        ros2_node->get_node_graph_interface()->get_node_names_and_namespaces();
+      for (auto & pair : node_names_and_namespaces) {
+        if (pair.first == ros2_node->get_name() && pair.second == ros2_node->get_namespace()) {
+          continue;
+        }
+        std::map<std::string, std::vector<std::string>> services_and_types =
+          ros2_node->get_service_names_and_types_by_node(pair.first, pair.second);
+        for (auto & it : services_and_types) {
+          service_names.insert(it.first);
+        }
+      }
+
       auto ros2_services_and_types = ros2_node->get_service_names_and_types();
       std::map<std::string, std::map<std::string, std::string>> active_ros2_services;
       for (const auto & service_and_types : ros2_services_and_types) {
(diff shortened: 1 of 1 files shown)


src/dynamic_bridge.cpp
```diff
@@ -725,12 +740,14 @@ int main(int argc, char * argv[])
           fprintf(stderr, "invalid service type '%s', skipping...\n", service_type.c_str());
           continue;
         }
-        auto service_type_package_name = service_type.substr(0, separator_position);
-        auto service_type_srv_name = service_type.substr(separator_position + 1);
 
-        // TODO(wjwwood): fix bug where just a ros2 client will cause a ros1 service to be made
```
dirk-thomas, line 731: The patch also fixes this issue.

PR #268, opened 2020-05-25
When generating service mappings cast pair to set to avoid duplicate pairs

suddrey-qut:
Fixes issue #257

Recreation of PR #259 to do master->master

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -562,17 +562,23 @@ def determine_common_services(
     for rule in mapping_rules:
         for ros1_srv in ros1_srvs:
             for ros2_srv in ros2_srvs:
+
+                pair = (ros1_srv, ros2_srv)
+
+                if pair in pairs:
+                    continue
+
                 if rule.ros1_package_name == ros1_srv.package_name and \
                    rule.ros2_package_name == ros2_srv.package_name:
                     if rule.ros1_service_name is None and rule.ros2_service_name is None:
                         if ros1_srv.message_name == ros2_srv.message_name:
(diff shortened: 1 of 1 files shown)


ros1_bridge/__init__.py
```diff
@@ -562,17 +562,23 @@ def determine_common_services(
     for rule in mapping_rules:
         for ros1_srv in ros1_srvs:
             for ros2_srv in ros2_srvs:
+
+                pair = (ros1_srv, ros2_srv)
+
+                if pair in pairs:
+                    continue
+
```
peterpena, line 570: nitpick: eliminate the newlines in line 565, 567, 570. Otherwise LGTM.

suddrey-qut: Done

dirk-thomas:
Thanks for the patch.

peterpena:
Should this be backported?

dirk-thomas:
> Should this be backported?

Into which ROS distro?

peterpena:
@dirk-thomas I am not sure what ROS distro it will be. The bug seems to be introduced all the way back to Ardent. What is the protocol to decide the ROS distro?

dirk-thomas:
> What is the protocol to decide the ROS distro?

Only active distro can be patched (so the earliest would be Dashing). Another criteria is if it is necessary / severe enough. Based on the original issue it seems trivial to avoid this case by not defining a custom mapping which replicates the default behavior. So I would think the need to backport is fairly low.

peterpena:
I see. Thanks for the explanation. I felt the need to ask if it needed to be backported, but based on what you said, it seems that it might not be necessary.

PR #272, opened 2020-06-08
fix multiple definition if message with same name as service exists

dirk-thomas:
Fixes the problem reported in https://answers.ros.org/question/354302/ros1_bridge-build-error-for-turtlebot3/

CI build just to check for regressions: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/371/)


dirk-thomas:
Fast-forwarded the `foxy` branch.

PR #282, opened 2020-08-19
use hardcoded QoS (keep all, transient local) for /tf_static topic in dynamic_bridge

dirk-thomas:
Otherwise the latched ROS 1 message won't be "latched" on the ROS 2 side and commonly ROS 2 subscribers of the `/tf_static` topic will use the QoS `transient_local` and therefore won't be matched with `volatile` publishers.

[Build Status](https://ci.ros2.org/job/ci_packaging_linux/415/)

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -112,8 +112,7 @@ public:
     ros::Publisher ros1_pub,
     rclcpp::PublisherBase::SharedPtr ros2_pub = nullptr)
   {
-    auto qos = rclcpp::SensorDataQoS();
-    qos.keep_last(queue_size);
+    auto qos = rclcpp::SensorDataQoS(rclcpp::KeepLast(queue_size));
     return create_ros2_subscriber(node, topic_name, qos, ros1_pub, ros2_pub);
   }
 
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -112,9 +112,9 @@ public:
     ros::Publisher ros1_pub,
     rclcpp::PublisherBase::SharedPtr ros2_pub = nullptr)
   {
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
-    custom_qos_profile.depth = queue_size;
-    return create_ros2_subscriber(node, topic_name, custom_qos_profile, ros1_pub, ros2_pub);
+    auto qos = rclcpp::SensorDataQoS();
+    qos.keep_last(queue_size);
+    return create_ros2_subscriber(node, topic_name, qos, ros1_pub, ros2_pub);
   }
 
   rclcpp::SubscriptionBase::SharedPtr
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -70,6 +70,15 @@ public:
   {
     auto qos = rclcpp::QoS(rclcpp::KeepAll());
     qos.get_rmw_qos_profile() = qos_profile;
+    return create_ros2_publisher(node, topic_name, qos);
+  }
+
+  rclcpp::PublisherBase::SharedPtr
+  create_ros2_publisher(
+    rclcpp::Node::SharedPtr node,
+    const std::string & topic_name,
+    const rclcpp::QoS & qos)
+  {
     return node->create_publisher<ROS2_T>(topic_name, qos);
   }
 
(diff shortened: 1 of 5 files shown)


src/bridge.cpp
```diff
@@ -55,13 +77,42 @@ create_bridge_from_2_to_1(
   const std::string & ros1_topic_name,
   size_t publisher_queue_size,
   rclcpp::PublisherBase::SharedPtr ros2_pub)
+{
+  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
+  custom_qos_profile.depth = subscriber_queue_size;
+  auto subscriber_qos = rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(custom_qos_profile));
+  subscriber_qos.get_rmw_qos_profile() = custom_qos_profile;
```
wjwwood, line 84:
```suggestion
  auto subscriber_qos = rclcpp::SensorDataQoS(subscriber_queue_size);
```

dirk-thomas: Thanks - applied in 5b2dcabf6827dd8f27feae567ea5b7136db9e421.

dirk-thomas: It seems that `SensorDataQoS` doesn't have a constructor taking a queue size. I modified the line in 6c256577384324162d3f149d0e920d452c694a00.

I also applied the same change to existing code in the `factory.hpp`: e3ca885d990382f9614713ac2a2e873bc3ae4262.

include/ros1_bridge/factory.hpp
```diff
@@ -103,9 +112,9 @@ class Factory : public FactoryInterface
     ros::Publisher ros1_pub,
     rclcpp::PublisherBase::SharedPtr ros2_pub = nullptr)
   {
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
-    custom_qos_profile.depth = queue_size;
-    return create_ros2_subscriber(node, topic_name, custom_qos_profile, ros1_pub, ros2_pub);
+    auto qos = rclcpp::SensorDataQoS();
+    qos.keep_last(queue_size);
```
wjwwood, line 116: I don't think it's any better or worse, but you can do this I think (I thought the KeepLast constructor would work)

```suggestion
    auto qos = rclcpp::SensorDataQoS(rclcpp::KeepLast(queue_size));
```

dirk-thomas: Thanks, I updated both code locations with the recommended change: b5c4c5ccb98cd1ffc2499e341e8c90f2f67d4982.

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/416/)

vmanchal1:
Is this fix going to be available in any of the patches to Foxy?

wjwwood:
@vmanchal1 there was a new release recently, it should be in the next patch release of Foxy I guess: https://github.com/ros/rosdistro/pull/26380

dirk-thomas:
> there was a new release recently, it should be in the next patch release of Foxy I guess

I don't think that is the case. The patch has only landed on `master`. Neither has it been backported to `foxy` nor has the `foxy` branch been fast forwarded to the state of the `master` branch.

@jacobperron Are you ok with releasing new patch release into `rolling` as well as fast forwarding the `foxy` branch?

jacobperron:
> Are you ok with releasing new patch release into rolling as well as fast forwarding the foxy branch?

SGTM (I didn't look too closely at binary compatibility, but I'm not sure if we should worry about that for this leafy package).

dirk-thomas:
> I didn't look too closely at binary compatibility

This patch only adds API.

> Is this fix going to be available in any of the patches to Foxy?

I just made and upstream release of 0.9.4 on `master` and also fast forwarded `foxy`. The bloom release is still pending though since the release repo has some incompatible patches... @jacobperron ?

dirk-thomas:
The Foxy release: ros/rosdistro#26519.

dirk-thomas:
And the same release into Rolling: ros/rosdistro#26520.

royito55:
Is this change implemented in the `parameter_bridge`?

If it is not, does it mean I have to install it from source and make the changes to be able to recompile?
Or is there a way to edit the binary? (I know there probably isn't)

Finally, would this post address the issue of not all transforms appearing while using the `parameter_bridge`, even though they appear using the `dynamic_bridge`?
https://answers.ros.org/question/349230/problem-passing-tf_static-through-ros1_bridge-to-ros2/

Issue #284, opened 2020-09-07
Allow exclusion of / ignoring of msg types


norro:
Allow to exclude certain message types and/or packages from the rosbridge build. This could be done at cmake time, e.g., through cmake configuration options.

AFAIR, this can be helpful to:

- ignore message types that are not compatible between ROS 1 und ROS 2 or don't have a proper, therefore breaking the rosbridge build
- reduce rosbridge build time

norro:
I think both a blacklist as well as a whitelist of types might be reasonable.
What do you think?

clalancette:
Hi @norro ,

This sounds like a good enhancement to the bridge.  Could you create a pull request implementing this feature? Thanks.

norro:
Sure, I will give it a shot.

norro:
I think a *package* blacklist/whitelist is more suited for the use-case of reducing build time, whereas a *message* blacklist/whitelist is more suited for the use-case of ignoring message types that are not compatible between ROS 1 und ROS 2.

PR #294, opened 2020-11-17
[WiP] Package whitelist resp. blacklist

norro:
This PR introduces package whitelist resp. blacklist, see https://github.com/ros2/ros1_bridge/issues/284.

If a file `${CMAKE_SOURCE_DIR}/../../../BridgePackageInclude.List"` is found and non-empty (one ROS package name per line), only those ROS packages are mapped by the bridge. 
*Otherwise*, if a file `${CMAKE_SOURCE_DIR}/../../../BridgePackageIgnore.List"` is found and non-empty (one ROS package name per line), those ROS packages are ignored by the bridge.

This PR is very much work-in-progress with feedback requested, especially regarding the following points:
- User interface: Alternatives to looking for lists in the ROS workspace?
- Testing: No clever idea how to test this feature (needs to test different cmake setups?)
- General implementation

I consider *message* blacklist/whitelist as a next (and additional) step, after implementing this *package* blacklist/whitelist.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,9 @@ find_package(rclcpp REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(std_msgs REQUIRED)
 
+# include / ignore lists
+include(cmake/find_ignore_include_lists.cmake)
+
 # find ROS 1 packages
 set(cmake_extras_files cmake/find_ros1_package.cmake cmake/find_ros1_interface_packages.cmake)
 include(cmake/find_ros1_package.cmake)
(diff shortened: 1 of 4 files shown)


norro:
@clalancette I would very much appreciate your feedback.

sloretz:
Is the goal a feature that ignores ROS packages at compile time? If so, I would recommend a user interface of a CMake variable specified on the command line. When using `colcon` this would look like

```
colcon build --cmake-args -DROS1_BRIDGE_IGNORE_PKGS=std_msgs;std_srvs --packages-select ros1_bridge
```

norro:
@sloretz Yes, that's right.
I like your idea of using cmake variables and will adapt the PR accordingly soon.

Thanks for your feedback!

PR #256, opened 2020-05-05
Implemented action bridge

hsd-dev:
This is in continuation of the work we started last year (https://discourse.ros.org/t/introducing-action-bridge/9103). The PR maps the available actions on the path and generates the interface mappings. There's also an `action_bridge.cpp` which accepts the name and type of the action and creates an action bridge.  

**NOTE:** 
- The PR is for `eloquent` because I am having issues compiling the `master` branch (even without the action_bridge). Any help here welcome.
- I am having issues compiling the interface factory for `GripperCommand`, because the name of the action and the message within it is the same. It generates 2 files with identical function names (`first defined here`). 
```
control_msgs__msg__GripperCommand__factories.cpp
control_msgs__action__GripperCommand__factories.cpp
```
- `builtin_interfaces` seems to be ignored in case of services. I am not sure if [this](https://github.com/ipa-hsd/ros1_bridge/blob/cf3ad4561cb761c46314fcfc6e1f18d8c29e2af7/ros1_bridge/__init__.py#L851) is the best solution

```
if (ros1_type, ros2_type) not in message_string_pairs and not ros2_type.startswith("builtin_interfaces"):
```
- TODO: write tests, update code formatter in editor
- missing feature: auto-discovery

PR #295, opened 2020-11-18
remove duplicate services and msgs

v-lopez:
`iterate_packages` might return ROS2 paths for already found ROS1 pkgs 

This results in duplicated entries in `data` and double definitions in the generated cpp code.

This happens to me building for instance this package: https://github.com/pal-robotics/pal_msgs/tree/ros2-tests/pal_common_msgs

I am trying to have both ROS1 and ROS2 msgs on the same branch, and need to do some package.xml and CMakeLists hacking, which might be tricking `iterate_packages`.

I have 3 workspaces:

ros1_ws extending /opt/ros/noetic, built with caktin_make install
ros2_ws extending /opt/ros/foxy, built with colcon build
bridge_ws, extending ros1_ws and ros2_ws install dirs.

And `rosmsg.iterate_packages(rospack, rosmsg.MODE_MSG)` returns:
```
 ('pal_common_msgs', '/root/msgs/ros1_ws/install/share/pal_common_msgs/msg'),
 ('pal_common_msgs',
  '/root/msgs/ros2_ws/install/pal_common_msgs/share/pal_common_msgs/msg'),
```

Am I doing something wrong? I was surprised that this did not happen with other messages that are both installed in ROS1 and ROS2.


v-lopez:
I have a similar fix for actions for when https://github.com/ros2/ros1_bridge/pull/256 is merged, or if this is merged first I'll push there.

```
-        if match:
-            actions.append({
+        action_data = {
                 'ros1_name': pair[0].message_name,
                 'ros2_name': pair[1].message_name,
                 'ros1_package': pair[0].package_name,
                 'ros2_package': pair[1].package_name,
                 'fields': output
-            })
+            }
+        if match and action_data not in actions:
+            actions.append(action_data)
```

ros1_bridge/__init__.py
```diff
@@ -245,7 +245,10 @@ def get_ros1_messages(rospack=None):
     pkgs = sorted(x for x in rosmsg.iterate_packages(rospack, rosmsg.MODE_MSG))
     for package_name, path in pkgs:
         for message_name in rosmsg._list_types(path, 'msg', rosmsg.MODE_MSG):
-            msgs.append(Message(package_name, message_name, path))
+            # iterate_packages might return ROS2 paths for already found ROS1 pkgs
+            msg = Message(package_name, message_name, path)
+            if msg not in msgs:
+              msgs.append(msg)
```
sloretz, line 251: Is it guaranteed that the first message found will be the ROS 1 version and not the ROS 2 version? If not then this check might not be enough.

v-lopez: I don't think it's guaranteed, so you're right.

I've reimplented this, please take a look.

v-lopez:
Friendly ping, could we move this? Is it going to make it to foxy or  just galactic an newer?

Timple:
Probably not, galactic is EOL already...

Issue #187, opened 2019-04-18
use .idl files for the ROS 2 interfaces


dirk-thomas:
Otherwise interfaces with `.idl` extensions won't be picked up by the bridge.

ksuszka:
Isn't enough to change  `${${package_name}_INTERFACE_FILES}` in line https://github.com/ros2/ros1_bridge/blob/596eb40ff81bc0bed4badcd5dc301fe47955a4d5/CMakeLists.txt#L102 to `${${package_name}_IDL_FILES}` ? If I'm not mistaken, rosidl generates idl files for msg anyway, so there should be no change for msg files, but interfaces with only idl files can be supported.

At the moment we have this issue, that we cannot compile ros1_bridge if ros-dashing-autoware-auto-msgs package is installed as in the latest version it has only idl files and current ros1_bridge adds those messages to factory files but doesn't compile them so there are linker errors.

clalancette:
@wjwwood Is this solved by https://github.com/ros2/ros1_bridge/pull/296 ?

wjwwood:
> @wjwwood Is this solved by #296 ?

Yup!

@ksuszka it's quite a big more involved than that, since the object in Python you get from processing an msg file directly (which is what the `ros1_bridge` was doing) is different from the object you get from processing the idl file instead.

PR #296, opened 2020-11-18
[forward port] update to use rosidl_parser and .idl files rather than rosidl_adapter and .msg files

wjwwood:
This is forward porting a change from Dashing made in https://github.com/ros2/ros1_bridge/pull/292 into master.

CI:

- `ci_packaging_linux`: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/431/)

Once this one is merged, I'll backport it to Foxy and Eloquent.

closes https://github.com/ros2/ros1_bridge/issues/187

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ foreach(package_name ${ros2_interface_packages})
   if(NOT "${package_name}" STREQUAL "builtin_interfaces")
     list(APPEND generated_files "${generated_path}/${package_name}_factories.cpp")
     list(APPEND generated_files "${generated_path}/${package_name}_factories.hpp")
-    foreach(interface_file ${${package_name}_INTERFACE_FILES})
+    foreach(interface_file ${${package_name}_IDL_FILES})
       file(TO_CMAKE_PATH "${interface_file}" interface_name)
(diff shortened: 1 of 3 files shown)


wjwwood:
CI after fixing linting issue:

- [Build Status](https://ci.ros2.org/job/ci_packaging_linux/432/)

jacobperron (approved):
LGTM.

For backports, do you think we should address backwards compatibility concerns raised here https://github.com/ros2/ros1_bridge/pull/292#issuecomment-729889386 ?

wjwwood:
> For backports, do you think we should address backwards compatibility concerns raised here #292 (comment) ?

It seems like we're ok leaving Dashing as-is there, so I assume the same is true for backports to other branches (e.g. foxy, eloquent), does that sound right? @jacobperron @sloretz 

Either way I think this is the right way for this to be done on master.

jacobperron:
@wjwwood Friendly ping for a Foxy backport. I'm not sure if the ship has sailed for an Eloquent backport.

wjwwood:
Yeah, I was waiting for the fix in https://github.com/ros2/ros1_bridge/pull/298.

PR #297, opened 2020-11-23
very minor typo in comments

Vicidel:
as the documentation for parameter bridge is lacking, better to have comments without typo to avoid time waste on the user side

### DIFF
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -54,8 +54,8 @@ int main(int argc, char * argv[])
   const char * topics_parameter_name = "topics";
   // the services parameters need to be arrays
   // and each item needs to be a dictionary with the following keys;
-  // service: the name of the topic to bridge (e.g. '/service_name')
-  // type: the type of the topic to bridge (e.g. 'pkgname/srv/SrvName')
+  // service: the name of the service to bridge (e.g. '/service_name')
+  // type: the type of the service to bridge (e.g. 'pkgname/srv/SrvName')
   const char * services_1_to_2_parameter_name = "services_1_to_2";
   const char * services_2_to_1_parameter_name = "services_2_to_1";
   if (argc > 1) {
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -54,7 +54,7 @@ int main(int argc, char * argv[])
   const char * topics_parameter_name = "topics";
   // the services parameters need to be arrays
   // and each item needs to be a dictionary with the following keys;
-  // topic: the name of the topic to bridge (e.g. '/service_name')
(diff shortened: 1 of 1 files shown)


src/parameter_bridge.cpp
```diff
@@ -54,7 +54,7 @@ int main(int argc, char * argv[])
   const char * topics_parameter_name = "topics";
   // the services parameters need to be arrays
   // and each item needs to be a dictionary with the following keys;
-  // topic: the name of the topic to bridge (e.g. '/service_name')
+  // service: the name of the topic to bridge (e.g. '/service_name')
```
fujitatomoya, line 57:
```suggestion
  // service: the name of the service to bridge (e.g. '/service_name')
```

src/parameter_bridge.cpp
```diff
@@ -54,7 +54,7 @@ int main(int argc, char * argv[])
...
   // type: the type of the topic to bridge (e.g. 'pkgname/srv/SrvName')
```
fujitatomoya, line 58:
```suggestion
  // type: the type of the service to bridge (e.g. 'pkgname/srv/SrvName')
```

fujitatomoya (changes requested):

nuclearsandwich (approved):

nuclearsandwich:
@Vicidel these changes are probably small enough not to need a DCO but if you add one it will make the bot happy and we don't have to think about it.

Vicidel:
The DCO guide there doesn't work for me with error :
```
$ git push --force-with-lease origin patch-1
ERROR: Permission to ros2/ros1_bridge.git denied to Vicidel.
fatal: Could not read from remote repository.
```
I'll try to solve this for the future.

nuclearsandwich:
> The DCO guide there doesn't work for me with error :

I expect it is failing because I've applied the suggestions as a further commit on top of your changes. If you do a `git pull --rebase origin patch-1` before trying again I think that should do it.

sloretz:
@Vicidel friendly ping :) Thanks for the PR! Mind trying the fix @nuclearsandwich suggested?

Vicidel:
Sorry for the delay, I tested it but I still have errors:

`git pull --rebase origin patch-1
fatal: Couldn't find remote ref patch-1
`

`git push --force-with-lease origin patch-1
remote: Permission to ros2/ros1_bridge.git denied to Vicidel.
fatal: unable to access 'https://github.com/ros2/ros1_bridge.git/': The requested URL returned error: 403
`

jacobperron (approved):
Considering this is a small change to a comment, I'll just manually bypass the DCO bot in order to resolve this ticket.

Vicidel:
Sorry about this, I'm not used to open source and these kind of things. I will be careful in the future to comply to DCO.
Thank you for the merge

PR #298, opened 2020-12-05
fix bug with sizeof when type of the arrays differ

wjwwood:
This patch fixes a bug where the `sizeof` calls used to compare sizes of fixed sized arrays/sequences were faulty when elements of the arrays have different memory layouts.

We've tested it against the Apex use case that was failing to compile. It's ready for review, and then I'll work on backports.


jacobperron (approved):

wjwwood:
CI:

- Linux Packaging: [Build Status](https://ci.ros2.org/job/ci_packaging_linux/436/)

PR #301, opened 2020-12-08
[backport foxy] support idl files on the ROS 2 side of the bridge

wjwwood:
Backport of https://github.com/ros2/ros1_bridge/pull/296 and a subsequent bug fix #298.


wjwwood:
This technically breaks public API, but as discussed in this comment thread:

https://github.com/ros2/ros1_bridge/pull/292#issuecomment-729889386

We agreed that this was unlikely to be a public API in use outside of `ros1_bridge` and not worth the effort of a API compatible rework.

wjwwood:
CI:

- [Build Status](https://ci.ros2.org/job/ci_packaging_linux/438/)

jacobperron (approved):

wjwwood:
The compiler warning in the CI is from `rmw_fastrtps_shared_cpp`, so it's unrelated.

PR #303, opened 2020-12-21
Fix logging for updated rclcpp interface

mjcarroll:
This has caused failures in the nightly jobs on platforms where ros1_bridge is built.

I believe that it is related to this PR: https://github.com/ros2/rclcpp/pull/1442

@audrow Can you confirm that this was an expected side effect?

### DIFF
--- a/src/bridge.cpp
+++ b/src/bridge.cpp
@@ -125,7 +125,7 @@ create_bidirectional_bridge(
   const std::string & topic_name,
   size_t queue_size)
 {
-  RCLCPP_INFO(ros2_node->get_logger(), std::string("create bidirectional bridge for topic " + topic_name).c_str());
+  RCLCPP_INFO(ros2_node->get_logger(), ("create bidirectional bridge for topic " + topic_name).c_str());
   BridgeHandles handles;
   handles.bridge1to2 = create_bridge_from_1_to_2(
     ros1_node, ros2_node,
--- a/src/bridge.cpp
+++ b/src/bridge.cpp
@@ -125,7 +125,7 @@ create_bidirectional_bridge(
   const std::string & topic_name,
   size_t queue_size)
 {
-  RCLCPP_INFO(ros2_node->get_logger(), ("create bidirectional bridge for topic " + topic_name).c_str());
+  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic %s", topic_name.c_str());
   BridgeHandles handles;
   handles.bridge1to2 = create_bridge_from_1_to_2(
     ros1_node, ros2_node,
--- a/src/bridge.cpp
+++ b/src/bridge.cpp
@@ -125,7 +125,9 @@ create_bidirectional_bridge(
   const std::string & topic_name,
   size_t queue_size)
 {
-  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic %s", topic_name.c_str());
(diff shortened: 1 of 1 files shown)


src/bridge.cpp
```diff
@@ -125,7 +125,7 @@ create_bidirectional_bridge(
   const std::string & topic_name,
   size_t queue_size)
 {
-  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic " + topic_name);
+  RCLCPP_INFO(ros2_node->get_logger(), std::string("create bidirectional bridge for topic " + topic_name).c_str());
```
ivanpauno, line 128:
```suggestion
  RCLCPP_INFO(ros2_node->get_logger(), ("create bidirectional bridge for topic " + topic_name).c_str());
```

ivanpauno (approved):

src/bridge.cpp
```diff
@@ -125,7 +125,7 @@ create_bidirectional_bridge(
   const std::string & topic_name,
   size_t queue_size)
 {
-  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic " + topic_name);
+  RCLCPP_INFO(ros2_node->get_logger(), ("create bidirectional bridge for topic " + topic_name).c_str());
```
ivanpauno, line 128: Sorry, but now I remember that passing a "non-literal" string is a security issue, and that's why support for `std::string` as a format string was removed.

The strictly correct code would be:

```suggestion
  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic "%s", topic_name.c_str());
```

but I don't mind if we merge this as-is first.

audrow, line 128: Clang still may not like this, as you're passing in a nonliteral.

How about the following?
```
RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic %s", topic_name.c_str());
```
or
```
RCLCPP_INFO(ros2_node->get_logger(), "%s", ("create bidirectional bridge for topic " + topic_name).c_str());
```

audrow (commented):
I think Clang will raise a warning with the change that's been made. I've suggested two alternatives.

audrow:
Also, this is an expected side effect. After ros2/rclcpp#1442, the logging macros only accept `char *`s and the format argument should be a string literal.

src/bridge.cpp
mjcarroll: I think there is a stray quote in yours, so I went with: 

```
  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic %s", topic_name.c_str());
```

mjcarroll: I went with:
```
  RCLCPP_INFO(ros2_node->get_logger(), "create bidirectional bridge for topic %s", topic_name.c_str());
```

audrow (approved):
LGTM, with green CI!

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13311/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8242/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11042/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13352/)

mjcarroll:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/442/)

Issue #314, opened 2021-05-18
ROS1 Bridge doesn't work with Multithreaded Executor on a ROS2 service


joelbudu:
## Bug report

**Required Info:**

- Operating System: Ubuntu 18.04
  - 
- Installation type: Binary
  - 
- Version or commit hash: dashing
  - 
- DDS implementation: Default
  - 
- Client library (if applicable): N/A
  - 

#### Steps to reproduce issue

1. Create a service with a Multithreaded executor in ROS2
2. Connect to ROS1 via the ros1_bridge
3. Call the ROS2 service through the bridge.
Only a single thread runs on the serice

```

```

#### Expected behavior
Multithreaded executor should run multiple nodes in separate threads.
#### Actual behavior
ROS2 service is run in a single thread
#### Additional information

----
## Feature request
ros1_bridge should support nodes with the Multithreaded executor

iuhilnehc-ynos:
@joelbudu

> Create **a service** with a Multithreaded executor in ROS2
> Multithreaded executor should **run multiple nodes** in separate threads.

Do you expect that a service created by a node running on a multithreaded executor of ros2 can deal with multiple requests from ros1 at the same time? Or two more different services created by a node or nodes? I just want to make sure you make the right code (IIRC, the flag of Callback group, exclusive, will be checked inside if you use the default callback group only in one node) in ROS2.

About `ros1_bridge`, there are some locations that seem to be updated if we want to support dealing with multiple requests from ros1 at the same time. such as,

1. use CallbackGroup with `Reentrant` (It just bridge the data for the service, so I think we can use reentrant here.) for
https://github.com/ros2/ros1_bridge/blob/5d77ce0caae4133cbdafbfcfb89c91ba8d875771/include/ros1_bridge/factory.hpp#L354

2. use `MultipleThreadedExecutor` for
https://github.com/ros2/ros1_bridge/blob/5d77ce0caae4133cbdafbfcfb89c91ba8d875771/src/dynamic_bridge.cpp#L791

BTW: could you provide an example and more detailed steps to reproduce what you expect?

iuhilnehc-ynos:
Could you use `foxy` or `rolling`? Because it seems the [EOL](https://docs.ros.org/en/foxy/Releases.html) of `dashing` is on May, 2021.

joelbudu:
[...]

PR #315, opened 2021-05-24
support services run with multiple threads

iuhilnehc-ynos:
Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -328,7 +328,8 @@ public:
     ros::NodeHandle & ros1_node, rclcpp::Node::SharedPtr ros2_node, const std::string & name)
   {
     ServiceBridge1to2 bridge;
-    bridge.client = ros2_node->create_client<ROS2_T>(name);
+    bridge.group = ros2_node->create_callback_group(rclcpp::CallbackGroupType::Reentrant);
+    bridge.client = ros2_node->create_client<ROS2_T>(name, rmw_qos_profile_services_default, bridge.group);
     auto m = &ServiceFactory<ROS1_T, ROS2_T>::forward_1_to_2;
     auto f = std::bind(
       m, this, bridge.client, ros2_node->get_logger(), std::placeholders::_1,
@@ -351,7 +352,8 @@ public:
     f = std::bind(
       m, this, bridge.client, ros2_node->get_logger(), std::placeholders::_1,
       std::placeholders::_2, std::placeholders::_3);
-    bridge.server = ros2_node->create_service<ROS2_T>(name, f);
+    bridge.group = ros2_node->create_callback_group(rclcpp::CallbackGroupType::Reentrant);
+    bridge.server = ros2_node->create_service<ROS2_T>(name, f, rmw_qos_profile_services_default, bridge.group);
     return bridge;
   }
 
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -136,7 +136,8 @@ void update_bridge(
   std::map<std::string, Bridge2to1HandlesAndMessageTypes> & bridges_2to1,
   std::map<std::string, ros1_bridge::ServiceBridge1to2> & service_bridges_1_to_2,
   std::map<std::string, ros1_bridge::ServiceBridge2to1> & service_bridges_2_to_1,
-  bool bridge_all_1to2_topics, bool bridge_all_2to1_topics)
+  bool bridge_all_1to2_topics, bool bridge_all_2to1_topics,
+  std::vector<rclcpp::CallbackGroup::SharedPtr> & invalid_callback_group)
 {
   std::lock_guard<std::mutex> lock(g_bridge_mutex);
 
@@ -353,6 +354,8 @@ void update_bridge(
     if (ros1_services.find(it->first) == ros1_services.end()) {
       printf("Removed 2 to 1 bridge for service %s\n", it->first.data());
       try {
+        // store the callback group to ensure that it is deleted after the node
+        invalid_callback_group.push_back(it->second.group);
         it = service_bridges_2_to_1.erase(it);
       } catch (std::runtime_error & e) {
         fprintf(stderr, "There was an error while removing 2 to 1 bridge: %s\n", e.what());
@@ -368,6 +371,8 @@ void update_bridge(
       printf("Removed 1 to 2 bridge for service %s\n", it->first.data());
       try {
         it->second.server.shutdown();
+        // store the callback group to ensure that it is deleted after the node
+        invalid_callback_group.push_back(it->second.group);
         it = service_bridges_1_to_2.erase(it);
       } catch (std::runtime_error & e) {
         fprintf(stderr, "There was an error while removing 1 to 2 bridge: %s\n", e.what());
@@ -484,6 +489,7 @@ int main(int argc, char * argv[])
   std::map<std::string, Bridge2to1HandlesAndMessageTypes> bridges_2to1;
   std::map<std::string, ros1_bridge::ServiceBridge1to2> service_bridges_1_to_2;
   std::map<std::string, ros1_bridge::ServiceBridge2to1> service_bridges_2_to_1;
+  std::vector<rclcpp::CallbackGroup::SharedPtr> invalid_callback_group;
 
   // setup polling of ROS 1 master
   auto ros1_poll = [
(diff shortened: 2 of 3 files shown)


iuhilnehc-ynos:
This patch is to:

- make ros2 client call ros1 service via ros1_bridge at the same time if ros1 service is supported with multiple threads ([Multi-threaded Spinning](http://wiki.ros.org/roscpp/Overview/Callbacks%20and%20Spinning#Multi-threaded_Spinning)).

https://github.com/iuhilnehc-ynos/ros_tutorials/blob/44666088677028f4edeb6246b9f791c489dbb653/roscpp_tutorials/add_two_ints_server/add_two_ints_server.cpp#L57-L59

running `ros2 run examples_rclcpp_minimal_client client_main& ros2 run examples_rclcpp_minimal_client client_main` to expect this service can serve multiple requests at the same time.
```
[ INFO] [1621936604.891178820]: request: x=41, y=1
[ INFO] [1621936604.891784270]:   sending back response: [42]
[ INFO] [1621936604.892629346]: request: x=41, y=1
[ INFO] [1621936604.892643574]:   sending back response: [42]
              ^^^^^^^^^//\   the timestamp of two requests are almost same.
```

without this patch, the log as following,
```
[ INFO] [1621933196.546494184]: request: x=41, y=1
[ INFO] [1621933196.546513489]:   sending back response: [42]
[ INFO] [1621933198.556169433]: request: x=41, y=1
              ^^^^^^^^^^^^^^^^^^^^^/\   2 sec after to process the next request
[ INFO] [1621933198.556201921]:   sending back response: [42]
```

- make ros1 client call ros2 service via ros1_bridge at the same time if ros2 service is supported with multiple threads(MultiThreadExecutor and a callback with reentrant). 

https://github.com/iuhilnehc-ynos/ros2-examples/blob/0acb1b892f318ec7898b812d10765f538c93ebc7/rclcpp/services/minimal_service/main.cpp#L44-L46

running `rosservice call /add_two_ints '{a: 3, b: 4}'& rosservice call /add_two_ints '{a: 3, b: 4}'` expect to get the service log as following
```
[INFO] [1621936929.275491886] [minimal_service]: request: 3 + 4
[INFO] [1621936929.275632310] [minimal_service]: request: 3 + 4
```

without this patch, the log as following,
```
[INFO] [1621937408.339715576] [minimal_service]: request: 3 + 4
[INFO] [1621937410.340892268] [minimal_service]: request: 3 + 4
```

Notice that supporting the service with running multiple threads is for not only the one service(a custom callback group with reentrant set by user) but also multiple different services.

Could somebody help to review this PR?

iuhilnehc-ynos:
The failure about `The following signatures were invalid: EXPKEYSIG F42ED6FBAB17C654 Open Robotics <info@osrfoundation.org>` was mentioned on https://answers.ros.org/question/379190/apt-update-signatures-were-invalid-f42ed6fbab17c654/, I'll run `retest this please` after the CI is fixed.

include/ros1_bridge/factory.hpp
```diff
@@ -33,6 +34,34 @@
...
 {
 
+static rclcpp::CallbackGroup::SharedPtr get_callback_group(
+  rclcpp::Node::SharedPtr ros2_node,
+  const std::string & topic_name = "")
+{
+  auto node_base = ros2_node->get_node_base_interface();
+  auto group = node_base->get_default_callback_group();
+  if (topic_name.empty()) {
+    // create a callback group with Reentrant for creating ros2 clients and services
```
fujitatomoya, line 44: since this is a just bridge, it seems to be okay to have reentrant callback group for each client and service. that is the idea, correct? i just want to be sure if i understand this.

include/ros1_bridge/factory.hpp
```diff
@@ -33,6 +34,34 @@
...
+    static auto s_shared_callbackgroup =
+      ros2_node->create_callback_group(rclcpp::CallbackGroupType::Reentrant);
+    return s_shared_callbackgroup;
+  }
+
+  typedef std::map<std::string, rclcpp::CallbackGroup::SharedPtr> TopicCallbackGroupMap;
+  static TopicCallbackGroupMap s_topic_callbackgroups;
```
fujitatomoya, line 51: instead of introducing `static auto s_shared_callbackgroup`, can we have everything in this map including `rclcpp::CallbackGroup::SharedPtr` for service and client?

src/dynamic_bridge.cpp
```diff
@@ -784,11 +787,11 @@ int main(int argc, char * argv[])
 
 
   // ROS 1 asynchronous spinner
-  ros::AsyncSpinner async_spinner(1);
+  ros::AsyncSpinner async_spinner(0, &ros1_callback_queue);
```
fujitatomoya, line 790: this will use the all cores in the system to spin. i think using Single or Multi threaded executor is dependent on the use case. for example, someone just bridge the certain topic only, in that case we do not need Multi threaded executor.

i think at least we would want to provide the option for user, then user can decide what they need.

src/dynamic_bridge.cpp
```diff
@@ -784,11 +787,11 @@ int main(int argc, char * argv[])
...
   async_spinner.start();
 
   // ROS 2 spinning loop
-  rclcpp::executors::SingleThreadedExecutor executor;
+  rclcpp::executors::MultiThreadedExecutor executor;
```
fujitatomoya, line 794: same here.

include/ros1_bridge/factory.hpp
iuhilnehc-ynos: Yes, that is what I understood.

src/dynamic_bridge.cpp
iuhilnehc-ynos: OK, I'll introduce an option, and also append an argument (bool multithread = false) to some methods to check if it's necessary to create a custom `callbackgroup` or not.

include/ros1_bridge/factory.hpp
```diff
@@ -33,6 +34,34 @@
...
 
+static rclcpp::CallbackGroup::SharedPtr get_callback_group(
+  rclcpp::Node::SharedPtr ros2_node,
+  const std::string & topic_name = "")
+{
+  auto node_base = ros2_node->get_node_base_interface();
+  auto group = node_base->get_default_callback_group();
+  if (topic_name.empty()) {
+    // create a callback group with Reentrant for creating ros2 clients and services
+    static auto s_shared_callbackgroup =
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I want to reuse only one `callbackgroup` for all services and clients.

include/ros1_bridge/factory.hpp
iuhilnehc-ynos: Maybe I can just use the empty string as the key

src/dynamic_bridge.cpp
```diff
@@ -119,6 +123,7 @@ bool parse_command_options(
   bool bridge_all_topics = get_flag_option(args, "--bridge-all-topics");
   bridge_all_1to2_topics = bridge_all_topics || get_flag_option(args, "--bridge-all-1to2-topics");
   bridge_all_2to1_topics = bridge_all_topics || get_flag_option(args, "--bridge-all-2to1-topics");
+  multi_threads = get_flag_option(args, "--multi-threads");
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Yeah, maybe some guys want to set two different values for the threads count of ros1 and ros2 spinner.

fujitatomoya:
@clalancette @sloretz friendly ping.

fujitatomoya:
@gbiggs you might want to take a look, this fixes https://github.com/ros2/ros1_bridge/issues/314

joelbudu:
Any chance in this getting merged? @iuhilnehc-ynos @sloretz @clalancette

iuhilnehc-ynos:
@joelbudu

I am sorry about not figuring out an elegant fix for this. The answer is no for this PR, at least.

src/dynamic_bridge.cpp
```diff
@@ -776,22 +797,41 @@ int main(int argc, char * argv[])
...
+    std::chrono::seconds(1), [&ros2_poll, &check_ros1_flag] {
+      ros2_poll();
+      check_ros1_flag();
+    });
 
 
   // ROS 1 asynchronous spinner
-  ros::AsyncSpinner async_spinner(1);
-  async_spinner.start();
+  std::unique_ptr<ros::AsyncSpinner> async_spinner = nullptr;
```
meetgandhi-eic, line 818: May be we create one more thread here and use ros::MultiThreadedSpinner?

PR #322, opened 2021-08-06
Use FindPython3 and make Python dependency explicit

sloretz:
This makes `ros1_bridge` use `FindPython3`, and makes the dependency on Python 3 explicit

Related to ros2/python_cmake_module#6
Blocks ament/ament_cmake#355


sloretz:
CI LGTM and I checked it locally with Noetic/Rolling

nuclearsandwich (approved):

PR #325, opened 2021-08-26
Allow .msg/.srv files to be in subdirectories of msg/, srv/

janstrohbeck:
If one has the following ROS1 message layout:
```
msg/
  group1/
    MessageDefinition1.msg
    MessageDefinition2.msg
  group2/
    MessageDefinition3.msg
```
That is, the `.msg` files are not directly in the `msg` folder, but in subfolders. Then the bridge does not find the message files correctly. The same goes for files in subfolders in the `srv` directory. This PR expands the search for `.msg` and `.srv` files to subdirectories, to make it work.

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -28,6 +28,7 @@ from rosidl_cmake import expand_template
 import rosidl_parser.parser
 
 import yaml
+import glob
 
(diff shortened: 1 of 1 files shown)


gbiggs:
@janstrohbeck You wouldn't happen to have a test of a minimal reproducible example I can use to verify this works, would you?

Issue #316, opened 2021-06-01
Unable to remap Parameter Bridge namespace


robin-read:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04 running in [foxy-ros1-bridge Docker image](https://hub.docker.com/_/ros)
- Installation type:
  - Built from source
- Version or commit hash:
  - 8339080 / tag: 0.10.1
- DDS implementation:
  - Fast-RTPS (default in Docker image)
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
The ultimate goal is to be able to start multiple bridges that have different names, bridge different topics/services and each run on a different computer.

To reproduce, attempt to start a ```parameter_bridge``` node and remap the node namespace from the command line.

``` ros2 run ros1_bridge parameter_bridge --ros-args -r __ns:=bridge_1```

#### Expected behavior

Bridge starts with a different name and can be inspected via ROS 1 and ROS2:

```
ros2 node list
ros2 node info /bridge_1/ros_bridge
```

```
rosnode list
rosnode info /bridge_1/ros_bridge`
```

#### Actual behavior

Node crashes with the following error message:

```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLInvalidROSArgsError'
  what():  failed to initialize rcl: Couldn't parse trailing -r flag. No remap rule found., at /tmp/binarydeb/ros-foxy-rcl-1.1.11/src/rcl/arguments.c:357
```

iuhilnehc-ynos:
[Supporting: Change the Default Namespace](https://design.ros2.org/articles/static_remapping.html)

`The replacement side must have a FQN with no special operators. `

Even if we use a FQN for ros2 namespace, there still exists an error. Because [ros::init](https://github.com/ros/ros_comm/blob/f5fa3a168760d62e9693f10dcb9adfffc6132d22/clients/roscpp/src/libros/init.cpp#L491-L510), which update the `arg(c|v)` if `:=` exists, is called before `rclcpp::init`.

Solution:
[...]

PR #339, opened 2021-12-13
change order in node initialization

sgermanserrano:
Follows [comment](https://github.com/ros2/ros1_bridge/issues/316#issuecomment-853571152) in #316 

Signed-off-by: Servando German Serrano <servando.german.serrano@gmail.com>

### DIFF
--- a/src/parameter_bridge.cpp
+++ b/src/parameter_bridge.cpp
@@ -33,13 +33,13 @@
 
 int main(int argc, char * argv[])
 {
-  // ROS 1 node
-  ros::init(argc, argv, "ros_bridge");
-  ros::NodeHandle ros1_node;
-
   // ROS 2 node
   rclcpp::init(argc, argv);
   auto ros2_node = rclcpp::Node::make_shared("ros_bridge");
+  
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
a couple of comments.

- Is this modification only required to `parameter_bridge`? probably we need to apply this fix to other places such as `simple_bridge`?
- it would be nice to add test to make sure.

src/parameter_bridge.cpp
```diff
@@ -33,13 +33,13 @@
...
 {
-  // ROS 1 node
-  ros::init(argc, argv, "ros_bridge");
-  ros::NodeHandle ros1_node;
-
   // ROS 2 node
   rclcpp::init(argc, argv);
   auto ros2_node = rclcpp::Node::make_shared("ros_bridge");
+  
+  // ROS 1 node
```
fujitatomoya, line 40: perhaps we would want to add comments why this `rclcpp::init` needs to be called before `ros::init`.

fujitatomoya:
CC: @iuhilnehc-ynos

PR #340, opened 2021-12-16
Parametrizing service execution timeout

nirwester:
By convention, service calls in ROS should have a short execution time. However, since the bridge currently doesn't yet have full support for actions, services are sometimes used where a bit of computation time is needed.

Currently the service call timeout on service_bridge_1_to_2 is hard-coded to 5 seconds for all services. This constrain is not transparent to the user, that needs to take a look at the code to understand what's going on.

The PR adds a new parameter, "service_execution_timeout", that can be set by the user to extend (or reduce) the required timeout for service calls. This was added for both the dynamic_bridge and the parameter_bridge.

### DIFF
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -327,8 +327,9 @@ void update_bridge(
     }
   }
 
-  int service_execution_timeout;
-  ros1_node.param<int>("service_execution_timeout", service_execution_timeout, 5);
+  int service_execution_timeout{5};
+  ros1_node.getParamCached(
+    "ros1_bridge/dynamic_bridge/service_execution_timeout", service_execution_timeout);
 
   // create bridges for ros2 services
   for (auto & service : ros2_services) {
--- a/src/dynamic_bridge.cpp
+++ b/src/dynamic_bridge.cpp
@@ -327,6 +327,9 @@ void update_bridge(
     }
   }
 
+  int service_execution_timeout;
+  ros1_node.param<int>("service_execution_timeout", service_execution_timeout, 5);
+
   // create bridges for ros2 services
   for (auto & service : ros2_services) {
     auto & name = service.first;
@@ -339,7 +342,8 @@ void update_bridge(
         "ros2", details.at("package"), details.at("name"));
       if (factory) {
         try {
-          service_bridges_1_to_2[name] = factory->service_bridge_1_to_2(ros1_node, ros2_node, name);
+          service_bridges_1_to_2[name] = factory->service_bridge_1_to_2(
+            ros1_node, ros2_node, name, service_execution_timeout);
           printf("Created 1 to 2 bridge for service %s\n", name.data());
         } catch (std::runtime_error & e) {
           fprintf(stderr, "Failed to created a bridge: %s\n", e.what());
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -294,7 +294,8 @@ public:
 
   bool forward_1_to_2(
     rclcpp::ClientBase::SharedPtr cli, rclcpp::Logger logger,
-    const ROS1Request & request1, ROS1Response & response1)
+    const ROS1Request & request1, ROS1Response & response1,
+    int service_execution_timeout)
   {
     auto client = std::dynamic_pointer_cast<rclcpp::Client<ROS2_T>>(cli);
     if (!client) {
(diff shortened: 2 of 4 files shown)


src/dynamic_bridge.cpp
```diff
@@ -327,6 +327,9 @@ void update_bridge(
     }
   }
 
+  int service_execution_timeout;
+  ros1_node.param<int>("service_execution_timeout", service_execution_timeout, 5);
```
fujitatomoya, line 331: i guess it is okay to have the common parameter for all bridges. but if we want to, we can use dedicated parameter for each bridge types.

e.g) `ros1_bridge/dynamic_bridge/service_execution_timeout` and `ros1_bridge/parameter_bridge/service_execution_timeout` ?

fujitatomoya: we could use `getParamCached` to process this faster, so that we do not query the parameter to rosmaster every single second.

fujitatomoya (commented):
this sounds reasonable for me, just a couple of minor comments.

src/dynamic_bridge.cpp
nirwester: Makes sense, updated :)

gbiggs (approved):

gbiggs:
LGTM pending CI.

nirwester:
Made it Ament-friendly to make CI happy

fujitatomoya:
@gbiggs @sloretz can you run CI?

gbiggs:
Gist: https://gist.github.com/gbiggs/d25e3d7ba7370e48c391d1dcf73b8815
BUILD args: --packages-up-to ros1_bridge
TEST args: --packages-select ros1_bridge
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9593/

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/15874/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/10551/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/13540/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/16177/)

fujitatomoya:
CI is failing cz of https://ci.ros2.org/job/ci_linux/15874/console

```
14:22:52 Package 'ros1_bridge' specified with --packages-up-to was not found
```

i am not sure what went wrong, and cannot reproduce this problem either in my local environment. any idea?

nirwester:
I'm not familiar with this integration process, but having ros1_bridge in the packages to ignore
23:22:52 Create marker file: src/ros2/ros1_bridge/COLCON_IGNORE
Looks fishy... how are these ignore packages chosen?

clalancette:
Running CI for the ros1_bridge is slightly tricky.  We don't support the bridge on anything but Linux, so you don't need to run CI for Windows or macOS.

For Linux, the easiest way to run CI with the ros1_bridge is to run the ci_packaging jobs:

* https://ci.ros2.org/job/ci_packaging_linux
* https://ci.ros2.org/job/ci_packaging_linux-aarch64

If you put your .repos file into the parameters there, then it should do the correct thing while trying to build the bridge.

fujitatomoya:
@clalancette thanks!

CI for linux and linux-aarch64:
- https://ci.ros2.org/job/ci_packaging_linux/469/
- https://ci.ros2.org/job/ci_packaging_linux-aarch64/74/

gbiggs:
Linux CI looks good, but aarch64 is failing with a strange error in a Python module. @clalancette Is it expected that aarch64 will fail for `ros1_bridge`?

clalancette:
> Linux CI looks good, but aarch64 is failing with a strange error in a Python module. @clalancette Is it expected that aarch64 will fail for ros1_bridge?

No, it should work.  But I think we were having some unrelated issues.  Here's another run of CI:

- Linux [Build Status](https://ci.ros2.org/job/ci_packaging_linux/473/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_packaging_linux-aarch64/84/)

gbiggs:
The instability looks to be in `ros1_bridge`. I think that's safe to ignore, so this can be merged?

clalancette:
The linting failures are almost certainly because this PR would need to be rebased onto the latest.  That we can ignore for now.  This otherwise looks good, so I'm going to go ahead and merge it.  Thanks for the contribution.

PR #342, opened 2022-02-03
Cleanup of README.md

ckaran:
I noticed that the README.md file had a number of minor issues that meant that it wasn't as clear as it could be.  This PR clears them up.

### Issues

- I rewrapped all lines to 80 columns.  This makes it a little easier to read in a text editor.
- I added `bash` as the [info string](https://github.github.com/gfm/#info-string) of every fenced code block.  This makes Github render the blocks correctly, and makes copying simpler.
- I added two new environment variables to the prerequisites.  These are `ROS1_INSTALL_PATH`, and `ROS2_INSTALL_PATH`.  They exist to make it simple to run the scripts regardless of which versions of ROS 1 and 2 are currently in use.  Instructions on how to set them are given within the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ These are defined as the paths to the installation location of their respective 
 If you installed Noetic in the default location, then the definition of `ROS1_INSTALL_PATH` will be `/opt/ros/noetic`.
 
 Building the bridge as described below requires you to build all of ROS 2.
-We assume that you have downloaded it to `~/ros2_galactic`, and that is where you plan on building it.
-In this case, `ROS2_INSTALL_PATH` will be defined as `~/ros2_galactic/install`.
+We assume that you have downloaded it to `~/ros2_rolling`, and that is where you plan on building it.
+In this case, `ROS2_INSTALL_PATH` will be defined as `~/ros2_rolling/install`.
 
 If you've chosen to install either or both versions of ROS somewhere else, you will need adjust the definitions below to match your installation paths.
 
@@ -63,7 +63,7 @@ Modify these definitions as appropriate for the versions of ROS that you're usin
 
 ```bash
 export ROS1_INSTALL_PATH=/opt/ros/noetic
-export ROS2_INSTALL_PATH=~/ros2_galactic/install
+export ROS2_INSTALL_PATH=~/ros2_rolling/install
 ```
 
 Note that no trailing '/' character is used in either definition.
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@ Building the bridge as described below requires you to build all of ROS 2. I ass
 
 If you've chosen to install either or both versions of ROS somewhere else, you will need adjust the definitions below to match your installation paths.
 
-Because these definitions are used continuously throughout this file, it would be useful to copy the following lines to your `.bashrc` file (if you're using `bash` as your shell).  Modify these definitions as appropriate for the versions of ROS that you're using, and for the shell that you're using.
+Because these definitions are used continuously throughout this file, it would be useful to copy the following lines to your `.bashrc` file (if you're using `bash` as your shell).
+Modify these definitions as appropriate for the versions of ROS that you're using, and for the shell that you're using.
 
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ We don't recommend having your ROS 1 environment sourced during this step as it 
 colcon build --symlink-install --packages-skip ros1_bridge
 ```
 
-Next you need to source the ROS 1 environment.  If you set up the `ROS1_INSTALL_PATH` environment variable as described earlier, then the following will source the correct `setup.bash` file.
+Next you need to source the ROS 1 environment.
+If you set the `ROS1_INSTALL_PATH` environment variable as described above, then the following will source the correct `setup.bash` file.
 
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ arguments. We don't recommend having your ROS 1 environment sourced during this
 step as it can add other libraries to the path.
 
 
-```
+```bash
 colcon build --symlink-install --packages-skip ros1_bridge
 ```
 
@@ -89,7 +89,7 @@ Next you need to source the ROS 1 environment, for Linux and ROS Melodic that
 would be:
 
 
-```
+```bash
 source /opt/ros/melodic/setup.bash
 # Or, on OSX, something like:
 # . ~/ros_catkin_ws/install_isolated/setup.bash
@@ -105,7 +105,7 @@ it builds the bridge. Alternatively you can do it manually by sourcing the
 relevant workspaces yourself, e.g.:
 
 
-```
+```bash
 # You have already sourced your ROS installation.
 # Source your ROS 2 installation:
 . <install-space-with-ros2>/local_setup.bash
@@ -118,7 +118,7 @@ relevant workspaces yourself, e.g.:
 Then build just the ROS 1 bridge:
 
 
-```
+```bash
 colcon build --symlink-install --packages-select ros1_bridge --cmake-force-configure
 ```
 
@@ -142,7 +142,7 @@ in their environment.
 First we start a ROS 1 `roscore`:
 
 
-```
+```bash
 # Shell A (ROS 1 only):
 . /opt/ros/melodic/setup.bash
 # Or, on OSX, something like:
@@ -157,7 +157,7 @@ topics. Once a *matching* topic has been detected it starts to bridge the
 messages on this topic.
 
 
-```
+```bash
 # Shell B (ROS 1 + ROS 2):
 # Source ROS 1 first:
 . /opt/ros/melodic/setup.bash
@@ -179,7 +179,7 @@ ROS 2 in a regular interval.
 Now we start the ROS 1 talker.
 
 
-```
+```bash
 # Shell C:
 . /opt/ros/melodic/setup.bash
 # Or, on OSX, something like:
@@ -194,7 +194,7 @@ The ROS 1 node will start printing the published messages to the console.
 Now we start the ROS 2 listener from the `demo_nodes_cpp` ROS 2 package.
 
 
-```
+```bash
 # Shell D:
 . <install-space-with-ros2>/setup.bash
 ros2 run demo_nodes_cpp listener
@@ -206,7 +206,7 @@ When looking at the output in *shell B* there will be a line stating that the
 bridge for this topic has been created:
 
 
-```
+```bash
 created 1to2 bridge for topic '/chatter' with ROS 1 type 'std_msgs/String' and ROS 2 type 'std_msgs/String'
 ```
 
@@ -215,7 +215,7 @@ the listener in *shell B* a line will be stating that the bridge has been torn
 down:
 
 
-```
+```bash
 removed 1to2 bridge for topic '/chatter'
 ```
 
@@ -230,7 +230,7 @@ The steps are very similar to the previous example and therefore only the
 commands are described.
 
 
-```
+```bash
 # Shell A:
 . /opt/ros/melodic/setup.bash
 # Or, on OSX, something like:
@@ -241,7 +241,7 @@ roscore
 ---
 
 
-```
+```bash
 # Shell B:
 . /opt/ros/melodic/setup.bash
 # Or, on OSX, something like:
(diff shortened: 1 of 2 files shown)


gbiggs:
Text (and similar, e.g. Markdown) files in ROS 2 are intentionally not wrapped. We use one-sentence-per-line because it makes diffs in PRs much easier to handle.

ckaran:
Understood.  I'll unwrap everything and resubmit.

ckaran:
@gbiggs I've fixed the wrapping as requested.

README.md
```diff
@@ -44,6 +44,26 @@ To run the following examples you will also need these ROS 1 packages:
...
+Building the bridge as described below requires you to build all of ROS 2. I assume that you have downloaded it to `~/ros2_galactic`, and that is where you plan on building it. In this case, `ROS2_INSTALL_PATH` will be defined as `~/ros2_galactic/install`.
+
+If you've chosen to install either or both versions of ROS somewhere else, you will need adjust the definitions below to match your installation paths.
+
+Because these definitions are used continuously throughout this file, it would be useful to copy the following lines to your `.bashrc` file (if you're using `bash` as your shell).  Modify these definitions as appropriate for the versions of ROS that you're using, and for the shell that you're using.
+
+
+```bash
+export ROS1_INSTALL_PATH=/opt/ros/noetic
+export ROS2_INSTALL_PATH=~/ros2_galactic/install
```
Timple, line 62: I'd say you have them both from binaries (building the standard bridge) or both from a workspace (building custom messages).

I cannot say what the usecase would be from sourcing ROS1 from binaries and ROS2 from source.

ckaran: Doing this is *literally* what the instructions in the same README.md says for us to do; apparently this is what is needed to build the bridge.  The only reason for these settings is because of those instructions.

Timple: Well, in that case the comment is not directed to you but the original author :slightly_smiling_face: 

But you can ignore this then I guess.

ckaran: No problem, I understand the confusion!  When I was first going through the complete README to get the bridge working on my machine, my first thought was that I had misread things.  Turns out that they're completely accurate! :man_shrugging:

Timple: > Turns out that they're completely accurate! man_shrugging

I didn't say that. I still think they are wrong. It's just not your doing ;-)

---
And I'm not a maintainer here, just an interested bystander.

README.md
```diff
@@ -44,6 +44,26 @@ To run the following examples you will also need these ROS 1 packages:
 * `rostopic`
 * `rqt_image_view`
 
+### Prerequisites for the examples in this file
+
+In order to make the examples below portable between versions of ROS, we've chosen to define two environment variables, `ROS1_INSTALL_PATH` and `ROS2_INSTALL_PATH`.  These are defined as the paths to the installation location of their respective ROS versions.
```
gbiggs, line 49:
```suggestion
In order to make the examples below portable between versions of ROS, we've chosen to define two environment variables, `ROS1_INSTALL_PATH` and `ROS2_INSTALL_PATH`. 
These are defined as the paths to the installation location of their respective ROS versions.
```

README.md
```diff
@@ -44,6 +44,26 @@ To run the following examples you will also need these ROS 1 packages:
...
+
+If you installed Noetic in the default location, then the definition of `ROS1_INSTALL_PATH` would likely be `/opt/ros/noetic`.
+
+Building the bridge as described below requires you to build all of ROS 2. I assume that you have downloaded it to `~/ros2_galactic`, and that is where you plan on building it. In this case, `ROS2_INSTALL_PATH` will be defined as `~/ros2_galactic/install`.
```
gbiggs, line 53:
```suggestion
Building the bridge as described below requires you to build all of ROS 2.
We assume that you have downloaded it to `~/ros2_galactic`, and that is where you plan on building it.
In this case, `ROS2_INSTALL_PATH` will be defined as `~/ros2_galactic/install`.
```

README.md
```diff
@@ -44,6 +44,26 @@ To run the following examples you will also need these ROS 1 packages:
...
+
+If you've chosen to install either or both versions of ROS somewhere else, you will need adjust the definitions below to match your installation paths.
+
+Because these definitions are used continuously throughout this file, it would be useful to copy the following lines to your `.bashrc` file (if you're using `bash` as your shell).  Modify these definitions as appropriate for the versions of ROS that you're using, and for the shell that you're using.
```
gbiggs, line 57:
```suggestion
Because these definitions are used continuously throughout this file, it would be useful to copy the following lines to your `.bashrc` file (if you're using `bash` as your shell).
Modify these definitions as appropriate for the versions of ROS that you're using, and for the shell that you're using.
```

README.md
```diff
@@ -44,6 +44,26 @@ To run the following examples you will also need these ROS 1 packages:
...
+
+
+```bash
+export ROS1_INSTALL_PATH=/opt/ros/noetic
+export ROS2_INSTALL_PATH=~/ros2_galactic/install
+```
+
+As a note, there is no trailing '/' character in either definition.  If you have problems involving paths, please verify that you have the correct path to the installation location, and that you do not have a trailing '/' in either definition.
```
gbiggs, line 65:
```suggestion
As a note, there is no trailing '/' character in either definition.
If you have problems involving paths, please verify that you have the correct path to the installation location, and that you do not have a trailing '/' in either definition.
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
 * `rostopic`
 * `rqt_image_view`
 
+### Prerequisites for the examples in this file
+
+In order to make the examples below portable between versions of ROS, we've chosen to define two environment variables, `ROS1_INSTALL_PATH` and `ROS2_INSTALL_PATH`.
```
gbiggs, line 49:
```suggestion
In order to make the examples below portable between versions of ROS, we define two environment variables, `ROS1_INSTALL_PATH` and `ROS2_INSTALL_PATH`. 
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
 * `rostopic`
 * `rqt_image_view`
 
+### Prerequisites for the examples in this file
+
+In order to make the examples below portable between versions of ROS, we've chosen to define two environment variables, `ROS1_INSTALL_PATH` and `ROS2_INSTALL_PATH`. 
+These are defined as the paths to the installation location of their respective ROS versions.
+
+If you installed Noetic in the default location, then the definition of `ROS1_INSTALL_PATH` would likely be `/opt/ros/noetic`.
```
gbiggs, line 52:
```suggestion
If you installed Noetic in the default location, then the definition of `ROS1_INSTALL_PATH` will be `/opt/ros/noetic`.
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+
+Building the bridge as described below requires you to build all of ROS 2.
+We assume that you have downloaded it to `~/ros2_galactic`, and that is where you plan on building it.
```
gbiggs, line 55:
```suggestion
We assume that you have downloaded it to `~/ros2`, and that is where you plan on building it.
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+In this case, `ROS2_INSTALL_PATH` will be defined as `~/ros2_galactic/install`.
```
gbiggs, line 56:
```suggestion
In this case, `ROS2_INSTALL_PATH` should be defined as `~/ros2/install`.
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+
+If you've chosen to install either or both versions of ROS somewhere else, you will need adjust the definitions below to match your installation paths.
+
+Because these definitions are used continuously throughout this file, it would be useful to copy the following lines to your `.bashrc` file (if you're using `bash` as your shell).
```
gbiggs, line 60:
```suggestion
Because these definitions are used continuously throughout this page, it is useful to add the following lines to your shell startup file (`~/.bashrc` if you are using `bash`, `~/.zshrc` if you are using `zsh`).
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+Modify these definitions as appropriate for the versions of ROS that you're using, and for the shell that you're using.
```
gbiggs, line 61:
```suggestion
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+
+
+```bash
+export ROS1_INSTALL_PATH=/opt/ros/noetic
+export ROS2_INSTALL_PATH=~/ros2_galactic/install
```
gbiggs, line 66:
```suggestion
export ROS2_INSTALL_PATH=~/ros2/install
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+```
+
+As a note, there is no trailing '/' character in either definition.
```
gbiggs, line 69:
```suggestion
Note that no trailing '/' character is used in either definition.
```

README.md
```diff
@@ -44,6 +44,31 @@ To run the following examples you will also need these ROS 1 packages:
...
+If you have problems involving paths, please verify that you have the correct path to the installation location, and that you do not have a trailing '/' in either definition.
```
gbiggs, line 70:
```suggestion
```

README.md
```diff
@@ -60,14 +85,16 @@ Here are the steps for Linux and OSX.
...
 We don't recommend having your ROS 1 environment sourced during this step as it can add other libraries to the path.
 
-```
+
+```bash
 colcon build --symlink-install --packages-skip ros1_bridge
 ```
 
-Next you need to source the ROS 1 environment, for Linux and ROS Melodic that would be:
+Next you need to source the ROS 1 environment.  If you set up the `ROS1_INSTALL_PATH` environment variable as described earlier, then the following will source the correct `setup.bash` file.
```
gbiggs, line 93:
```suggestion
Next you need to source the ROS 1 environment.
If you set the `ROS1_INSTALL_PATH` environment variable as described above, then the following will source the correct `setup.bash` file.
```

README.md
ckaran: What about mirroring what happens in `/opt/ros`?  That is, instead of `~/ros2_galactic`, the download location is `~/ros2/galactic`?  This lets end users download and install multiple versions of ROS 2 to their machines.

README.md
ckaran: I can change 'will' -> 'should', but I'd like to use the path suggestion from above.  Would that work for you?

README.md
ckaran: I don't want to remove this line for two reasons:

- While `bash` and `zsh` are common because they are the defaults on Linux and OS X respectively, they aren't the only available shells.  I want to remind users that they need to think about their shells.
- When I first started working on this update, it talked about `melodic` and `foxy`.  That suggests that this documentation may live long enough that the latest and greatest version of ROS 2 will change.  I want to make sure that we're letting end users know that they need to think about what is going on behind the scenes, rather than blindly following the instructions.

README.md
ckaran: See my earlier suggestion about the paths.

README.md
ckaran: I don't want to remove this.  It will help others when they are debugging what went wrong.

README.md
gbiggs: Upon further reflection, and checking the existing documentation, we should keep the path as `~/ros2_galactic` because that is where the installation documentation says to put it.

README.md
gbiggs: We have specific versions of the README for each ROS version. Which reminds me that this particular version, being `master`, should actually refer to ROS 2 Rolling and Noetic. The correct path is `~/ros2_rolling` and there is no need to instruct the user to alter the instructions for different ROS 2 versions.

README.md
ckaran: OK, I'll leave it as-is.  Thank you.

README.md
ckaran: Understood, I'll modify it.

ckaran:
OK, I modified 'galactic' -> 'rolling'.

gbiggs (approved):

gbiggs:
Thanks for the contribution!

ckaran:
You're welcome!

PR #351, opened 2022-03-31
Make sure to use fully-qualified boost placeholders.

clalancette:
Also make sure to declare a direct dependency on boost.
While we are always getting it from ROS 1, this code directly
uses it and so should also declare a dependency.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

I recently attempted to build the ros1_bridge against the ROS (1) packages available directly in Ubuntu Jammy.  This is the only thing that prevented me from doing so, and is an obvious fix anyway.


clalancette:
CI (packaging jobs, which build the bridge, and Focal, where we have Noetic support):

* Linux [Build Status](http://ci.ros2.org/job/ci_packaging_linux/496/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_packaging_linux-aarch64/96/)

PR #361, opened 2022-05-09
Update maintainers

quarkytale:
Added new maintainers and moved previous as authors!


clalancette (changes requested):
So in this particular case, you've been added as a maintainer, but we shouldn't remove the old maintainers.  There are enough issues with the bridge that I think warrants having multiple maintainers here, so please just add yourself and make no other changes.

clalancette (approved):
Looks good.

Since this is only adding something to the package.xml, once CI comes back green go ahead and merge it.

Issue #354, opened 2022-04-08
Fields selection


amifsud:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04

- Installation type: Binaries, ros1_bridge from sources
- Version or commit hash: Foxy HEAD
- DDS implementation: Compile time issue

#### Steps to reproduce issue

Trying to map the following messages in ros1:

Example.msg :
```
float64 a
float64 b
float64 c
float64 d
```
Example1.msg :
```
float64 a
geometry_msgs/Vector3 v
```
to those ones in ros2:

Example2.msg :
```
float64 w
float64 x
float64 y
float64 z
```
Example3.msg :
```
float64 w
geometry_msgs/Vector3 u
```

With the mapping rules :

```
-
    ros1_package_name: 'ros1_msgs'
    ros1_message_name: 'Example1'
    ros2_package_name: 'ros2_msgs'
    ros2_message_name: 'Example2'
        fields_1_to_2:
            a:   'w'
            v.x: 'x'
            v.y: 'y'
            v.z: 'z'
```

Or the following ones :
```
-
    ros1_package_name: 'ros1_msgs'
    ros1_message_name: 'Example'
    ros2_package_name: 'ros2_msgs'
    ros2_message_name: 'Example3'
        fields_1_to_2:
            a: 'w'
            b: 'u.x'
            c: 'u.y'
            d: 'u.z'
```

In ros1 workspace (~/ros1_ws):

```
source /opt/ros/noetic/setup.bash
catkin_make install

source ~/ros1_ws/install/setup.bash
rosmsg show ros1_msgs/Example
rosmsg show ros1_msgs/Example1
```

In ros2 workspace (~/ros2_ws):

```
source /opt/ros/foxy/setup.bash
colcon build --build-base foxy_build --install-base foxy_install

source ~/ros2_ws/foxy_install/setup.bash
ros2 interface show ros2_msgs/msg/Example2
ros2 interface show ros2_msgs/msg/Example3
```

In ros1_bridge workspace (~/bridge_ws):

```
source ~/ros1_ws/install/setup.bash
source ~/ros2_ws/foxy_install/setup.bash
colcon build --build-base foxy_build --install-base foxy_install --cmake-force-configure

source ~/bridge_ws/foxy_install/setup.bash
ros2 run ros1_bridge dynamic_bridge --print-pairs | grep Example
```

#### Expected behavior

In ros1 workspace (~/ros1_ws):

```
[...]

PR #366, opened 2022-06-09
Fixed index for ros2 namespace type messages

quarkytale:
As reported in #354 bridge fails when ros2 message uses standard namespaces (std_msgs, geometry_msgs, etc.) to be mapped with  ros1 messages. Note: this works as expected when no namespaces are used on ros2 side, regardless of what's on ros1 side.

It was failing while returning package name and message name parsed from a `rosidl_parser.definition.NamespacedType` object. Type field was incorrect, seems to be using `rosidl_parser` for messages and `rosidl_adapter` for services, as far as my understanding. Updated the function description too.

### DIFF
--- a/ros1_bridge/__init__.py
+++ b/ros1_bridge/__init__.py
@@ -959,8 +959,8 @@ class MessageIndex:
         """
         Get Message from ROS 2 index.
 
-        :type field: rosidl_adapter.parser.Field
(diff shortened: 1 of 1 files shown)


gbiggs (approved):

quarkytale:
CI_Linux_Packaging
 [Build Status](https://ci.ros2.org/job/packaging_linux/2673/)

PR #377, opened 2022-08-26
ros1_bridge function for converting generic message types

dbking77:
This adds functions to convert between generic message types for ROS1 (ShapeShifter) and ROS2 (SerializedMessage)

This is useful for custom bag recording tools. For us specifically, there is a tool (action_monitor) that records certain topics at specific times based on the current robot action. If robot is idle, we record very few topics. If robot is navigating, we record many more topics.
We'de like to extend this tool to record both ROS1 and ROS2 topics at the same time to a ROS1 bag format.
If we just used dynamic_bridge to bridge all ROS2 topics, we'de pay heavy penalty bridging unnecessary topics when robot is idle.

To subscribe to an arbitrary ROS2 topic, rclcpp provides create_generic_subscription which invokes a callback that takes SerializedMessage. Because there is no concrete type at compile time we cannot use convert_2_to_1.
Because each ros1_bridge::Factory is templated on both ROS1 and ROS2 message types, it has all the concrete type information to perform the generic message conversion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,29 @@ if(BUILD_TESTING)
     endif()
     set(TEST_ROS1_BRIDGE TRUE)
   endif()
+
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(geometry_msgs REQUIRED)
+  find_package(std_msgs REQUIRED)
+  find_package(sensor_msgs REQUIRED)
+  find_ros1_package(geometry_msgs)
+  find_ros1_package(std_msgs)
+  find_ros1_package(sensor_msgs)
+  if(ros1_geometry_msgs_FOUND AND ros1_sensor_msgs_FOUND AND ros1_std_msgs_FOUND)
+    ament_add_gtest(test_convert_generic test/test_convert_generic.cpp)
+    ament_target_dependencies(test_convert_generic
+      "ros1_roscpp"
+      "ros1_sensor_msgs"
+      "ros1_std_msgs"
+      "ros1_geometry_msgs"
+      "sensor_msgs"
+      "std_msgs"
+      "geometry_msgs"
+      "rclcpp"
+    )
+    target_link_libraries(test_convert_generic ${PROJECT_NAME})
+  endif()
+
 endif()
 
 ament_export_include_directories(include)
--- a/include/ros1_bridge/factory.hpp
+++ b/include/ros1_bridge/factory.hpp
@@ -19,6 +19,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "rmw/rmw.h"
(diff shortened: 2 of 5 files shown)


gbiggs:
Please update your PR's description with information about what new feature you are implementing and why. You also should add some tests of the new functionality.

dbking77:
@gbiggs  I added PR description and unit tests.  Could you approve the testing workflow?

src/shape_shifter_access.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "ros/ros.h"
+#include "ros/console.h"
+#include "ros/assert.h"
+#include <vector>
+#include <string>
+#include <string.h>
+#include <ros/message_traits.h>
+#include "topic_tools/macros.h"
+
+// Redefine the meaning of private for just new stuff in shape_shifter header
```
gbiggs, line 35: What's going on here?

dbking77: ShapeShifter does not allow access to its internal buffer (unlike the ROS2 SerializedMessage).   For the ROS1 use cases there is no need for this because functions are provided to serialize data from/to a ROS1 stream.   However, I want to convert the data stored inside a SerializedMessage into a concrete ROS2 message type without making another copy.

Since ROS1 is no longer maintained, it's not  possible to request a change to ROS that provides an accessor function to ShapeShifter's internal buffer.

Typically classes will make "internal" data protected/private when there is no obvious reason to allow access.   This provides flexibility to change in the internal representation in the future.  Since ROS1 is not getting any update, this no longer matters.

If this seems too ugly, I can always define another type like ShapeShifter that can act like a container for any generic ROS1 type.

dbking77: One thought I just had is I could just have generic function take a const reference to a buffer (std::vector<uint8_t>) and let external called figure out how they want get access to the buffer.
Like I said before it is possible for someone to create something very similar to the ShapeShifter to subscribe to "generic" ROS1 streams.

dbking77:
@gbiggs I replaced shape_shifter with a vector<uint8_t> buffer this does reduce dependencies and simplify stuff quite a bit.   Can I get a testing retrigger?

gbiggs:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17560/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12098/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18127/)

gbiggs (approved):

gbiggs:
Whoops, wrong CI configuration. Let's try that again.

[Build Status](https://ci.ros2.org/job/ci_packaging_linux/525/)

gbiggs:
Thanks for the contribution!

PR #384, opened 2022-11-05
Dockerfile and Docker compose

yaruchyo:
use Dockerfile to simply create a new container with ros_1 bridge

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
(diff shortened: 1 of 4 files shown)


methylDragon:
Several comments:

- I'm not sure I see the point of this PR, especially if the dockerfile is going to get pinned to a certain version of ROS.
- Furthermore the dockerfile isn't installed anyway, which means a user is required to have access to the source code to use the dockerfile, in which case, why not make it a README or doc section?
- Also, I think your IDE added a gitignore file that you might want to remove

methylDragon:
@yaruchyo 
I'm going to close this PR. If you'd like, you can make a new PR where the dockerfile is a part of the docs as an example usage, and we can re-review

:fire: :raised_hands:

PR #399, opened 2023-04-06
Improve bridge command parser

paulbovbel:
From https://github.com/ros2/ros1_bridge/pull/396


paulbovbel:
Wrong repository, please disregard

PR #402, opened 2023-05-05
Added support for custom parameter bridge node names

ayushghosh21:
In cases where you may need to run multiple parameter bridges at the same time (eg: in multiple docker containers connected to the same ROS1 master) I was unable to run due to bridge nodes containing the same name. Initially passing in a command to rename the ros2 bridge is not successful because ROS1 and ROS2 parse the command line arguments differently. 

Here, I am simply accounting for that, by reformatting the command line argument passed into ROS1 bridge node. This way I am able to specify custom names for the parameter bridge and therefore run multiple instances of the parameter bridge. 

You can now simply pass in the arguments for a custom node name for the parameter bridge:
```
ros2 run ros1_bridge parameter_bridge --ros-args -r __node:=my_custom_name
```


LucasHaug:
Heey @ayushghosh21, maybe it can help you, I opened the PR #396 some time ago and with the changes I did there you can change any ROS arg for the ROS 1 and the ROS 2 nodes, including the name, namespace and so on.

PR #401, opened 2023-05-04
Parametrization of `parameter_bridge` quality of service [Port of the commits (ec44770) and (86b4245) to foxy branch]

lucyannofrota:
Reproduces the changes made in commit https://github.com/ros2/ros1_bridge/pull/331 in ros foxy.

Parametrize Quality of Service in `parameter_bridge`

Signed-off-by: Lucyanno Frota <lucyannofrota@gmail.com>

PR #406, opened 2023-05-26
Remove xmlrpcpp from package.xml

quarkytale:
https://github.com/ros2/ros1_bridge/pull/401 added a dependency which should not be handled by rosdep, so removing from package.xml


quarkytale:
CI
 [Build Status](https://ci.ros2.org/job/ci_packaging_linux/546/)

Issue #393, opened 2023-01-18
[Humble] Building Failed


Ogunniran:
## Bug report

**Note:** It built properly after c2fa9ed11d1be6cb36441ad317d0797deb4714d9 was removed.

**Required Info:**

- Operating System:
  - Ubuntu 22.04
  - ROS1: Noetic
  - ROS2: Humble
- Installation type:
  - From source

Building the ros1_bridge with `colcon build --symlink-install --packages-select ros1_bridge --cmake-force-configure` failed with the error:

```
//home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp: In instantiation of ‘void ros1_bridge::streamPrimitiveVector(ros::serialization::OStream&, const VEC_PRIMITIVE_T&) [with VEC_PRIMITIVE_T = std::vector<bool, std::allocator<bool> >]’:
/home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp:830:24:   required from here
/home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp:694:46: error: lvalue required as unary ‘&’ operand
  694 |   memcpy(stream.advance(data_len), &vec.front(), data_len);
      |                                     ~~~~~~~~~^~
/home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp: In instantiation of ‘void ros1_bridge::streamPrimitiveVector(ros::serialization::IStream&, VEC_PRIMITIVE_T&) [with VEC_PRIMITIVE_T = std::vector<bool, std::allocator<bool> >]’:
/home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp:1063:24:   required from here
/home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp:711:20: error: taking address of rvalue [-fpermissive]
  711 |   memcpy(&vec.front(), stream.advance(data_len), data_len);
      |           ~~~~~~~~~^~
/home/x/ros1_bridge/ros_bridge/build/ros1_bridge/generated/...__factories.cpp:711:9: warning: ‘void* memcpy(void*, const void*, size_t)’ writing to an object of type ‘std::vector<bool, std::allocator<bool> >::reference’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
[...]

Issue #391, opened 2023-01-12
[Humble] No conversion from `rcl_interfaces/msg/Log` and `rosgraph_msgs/Log`


Timple:
## Bug report

**Required Info:**

- Operating System:
  - Docker (ubuntu:jammy)
- Installation type:
  - ROS1 from ubuntu binaries
  - ROS2 from source (because of catkin conflict)
  - `ros_bridge` from source
- Version or commit hash:
  - master
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - 

#### Steps to reproduce issue
```bash
git clone https://github.com/contradict/ros-humble-ros1-bridge.git
cd ros-humble-ros1-bridge
docker build . -t bridge
```

Then run with this docker-compose file:
```yaml
version: "2.1"
services:
  roscore:
    image: "ros:noetic"
    network_mode: host
    command: roscore
  ros1-bridge:
    image: bridge
    network_mode: host
    command: ros2 run ros1_bridge dynamic_bridge
```

#### Expected behavior
Bridging `/rosout`

#### Actual behavior
Not bridging `/rosout`

#### Additional information
Seems like the mapping is not found: https://github.com/ros2/rcl_interfaces/blob/humble/rcl_interfaces/mapping_rules.yaml

#### Relevant other issues:
https://github.com/ros2/ros1_bridge/issues/159 Was fixed, is broken again for humble
https://github.com/contradict/ros-humble-ros1-bridge/issues/2 (Not the fault of the dockerfile I think)

Timple:
PS. Also seems to be the case for rolling.

brow1633:
Not a fix, but a workaround to enable the use of ROS1 bridge with the above Dockerfile (without logging):

Instead of `roscore` run `rosmaster --core`

Timple:
But that would mean I would need to configure all nodes to run without `/rosout` right?

faraz7321:
Hi! Were you able to resolve the issue? I am also experiencing the same error.

brow1633:
I did something similar here, but with the addition of rosbridge_suite: https://github.com/brow1633/ros_docker/blob/main/ros-humble-bridge

I'm not sure about configuring other nodes.  I haven't run into issues, but haven't run this extensively to be honest.

Timple:
@brow1633 because you are using `rosmaster --core` instead of `roscore`.
Therefore you simply skip logging and there is no issue with converting the logging message.

Timple:
[...]

PR #411, opened 2023-07-07
Handle special case of std::vector<bool>

oKermorgant:
This PR modifies the `interface_factories` template file and adds a special case to handle messages which are `std:vector<bool>`.

The current approach to stream a `std::vector` is to `memcpy` its content into/from the stream.

Unfortunately, it does not work for `std::vector<bool>` as there is [not a single `bool`](https://stackoverflow.com/questions/17794569/why-isnt-vectorbool-a-stl-container) inside such.

In this PR the template file generates a special call when a `vector<bool>` is detected.

This should answer https://github.com/ros2/ros1_bridge/issues/391, https://github.com/ros2/ros1_bridge/issues/393


Timple:
@oKermorgant You probably should signoff on your commits: https://github.com/ros2/ros1_bridge/pull/411/checks?check_run_id=14855763467

oKermorgant:
The last push is to handle `vector<bool>` that lie in services, the code was generated elsewhere.
In order to handle this I had to use index-based for loops intead of iterators (you cannot iterate on a `vector<bool>` and hope to get a pointer to a bool...).

mikaelarguedas:
fixes the bool list issue for me :+1:  (setup: Ubuntu 24.04, ROS 2 Jazzy /  ROS Noetic)

Thanks @oKermorgant !

logic-bomb-gmr:
build fine for me as well, thanks @oKermorgant.
my message file:
```
.
.
.
bool[] data
.
.
.
```

setup: Ubuntu 22.04.5 LTS with ROS 2 Humble and ROS Noetic.

PR #424, opened 2023-10-24
Support tf_static 2to1

pawelirh:
## Changes

This PR allows to use of `dynamic_bridge` with `/tf_static` topic in the 2to1 direction of bridging.
Due to an invalid qos policy, the ROS2 subscriber created by the bridge was not able to receive the `tf_static` message. Furthermore, the ROS publisher wasn't latching the message.

- For tf_static bridging in 2to1 direction special rules are applied
- ROS2 subscriber special qos (reliable, keep all, transient local)
- ROS publisher latching enabled

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,8 @@
 Changelog for package ros1_bridge
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Forthcoming
------------
+0.9.7 (2023-05-27)
+------------------
 * remove xmlrpcpp from package.xml (`#406 <https://github.com/ros2/ros1_bridge/issues/406>`_)
 * Parametrization of `parameter_bridge` quality of service [Port of the commits (ec44770) and (86b4245) to foxy branch] (`#401 <https://github.com/ros2/ros1_bridge/issues/401>`_)
 * Contributors: Dharini Dutia, Lucyanno Frota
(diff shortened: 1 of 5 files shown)


weidinger-c:
Is this alreadyon the main branch for humble? I still cannot latch tf_static messages in ROS 1 Noetic.

pawelirh:
The master branch (same as `foxy`) has this rule applied in the ROS->ROS2 direction. This PR implements the corresponding change but allows to port of the `tf_static` message in the reverse direction ROS2->ROS too.

Feel free to use my fork until changes here are not merged. ->  https://github.com/hsrn24/ros1_bridge/tree/foxy
IMPORTANT: Changes are applied only in dynamic_bridge on a foxy branch

weidinger-c:
Thanks for your reply, but I am only using the parameter_bridge, because I only need a small subset of a large set of topics bridged. So I guess I need to wait for the fix on the parameter bridge.

pawelirh:
I don't think it's gonna happen shortly. 

However, the `parameter_bridge` allows for configuring the QoS. So theoretically you can achieve demanded behavior by doing so (I didn't try running parameter_bridge).

Did you check out this link?
https://github.com/ros2/ros1_bridge?tab=readme-ov-file#parametrizing-quality-of-service

Mat198:
I tested this PR with ROS 2 (Humble) <-> ROS 1 (Noetic) and it works correctly. Why is it not merged?

Alby407:
Tested on Jazzy, does not work.

PR #467, opened 2026-05-21
fix bug that when some fields are not existing in the ROS 2 message but the field is actually manually mapped, the whole mapping is incorrectly ignored

### LINKED ISSUE #391: [Humble] No conversion from `rcl_interfaces/msg/Log` and `rosgraph_msgs/Log`

fabianhirmann:
## Description

fix bug that when some fields are not existing in the ROS 2 message but the field is actually manually mapped, the whole mapping is incorrectly ignored

This at the end caused that for example the mapping between `rcl_interfaces/Log` and `rosgraph_msgs/Log` was no more created because it was incorrectly skipped due to `stamp` having a custom mapping and a corresponding field in the ROS 1 version with the same name does not exist (only header.stamp which is manually mapped).

Fixes #391 (at least on ROS 2 galactic but same issue probably occurs on ROS 2 Humble)

### Did you use Generative AI?

No

### Additional Information

The issue probably appeared after #382 was merged.


mjcarroll:
Just FYI, so you aren't waiting on a review here.

Since ROS 1 noetic went end of life, the ROS 2 PMC is no longer considering this a supported package, so we likely won't be taking pull requests against this package.

We will likely archive this repository or move it to some other maintainer who is interested in the near future.

fabianhirmann:
> Just FYI, so you aren't waiting on a review here.
> 
> Since ROS 1 noetic went end of life, the ROS 2 PMC is no longer considering this a supported package, so we likely won't be taking pull requests against this package.
> 
> We will likely archive this repository or move it to some other maintainer who is interested in the near future.

Thank you for the feedback @mjcarroll! I mainly opened this PR such that other users of the library with the same issue (like #391) are aware that there is a fix that one can merge into their fork. Since there isn't going on much in this repo, I anyway did not expect much chances of getting it merged these late days for the transition from ROS 1 to ROS 2. In any case again many thanks for the clear feedback!

